PR #27, opened 2014-02-06
New feature deployer

jjongsma:
Looking for some feedback on this one.

We do a lot of hot upgrades to our servers by just updating the local feature XML files, and have noticed some issues in the current version, most importantly:

1. Ref counting is broken, so features/bundles are frequently not uninstalled even when they are no longer required as a dependency of install=auto features
2. Removing or replacing a repository (feature XML file) does not remove bundles contained in it even if they were no longer referenced as a dependency anywhere, leaving a lot of running orphaned components

This feature deployer rewrite is our attempt to resolve these issues. It does change the behavior, but only to fix what currently seems broken.


jbonofre:
Most of this point should be now fixed on K4.

PR #48, opened 2014-10-04
highlight local classes/resources in "bundle:classes" command

fdie:


jbonofre:
Applied, can you close the pull request ? Thanks.

PR #51, opened 2014-11-06
[KARAF-3344] Fix karaf-maven-plugin non-standard URLs handling

boodoopl:
74515f50fa5c1f86c97a7c46688d253f7b2d60fd fixed only "wrap:" URL options handling.
I extended it to other URLs types and simplified implementation by using RegExp.


jbonofre:
As this PR is pretty old and can't be merged easily, and also as we already did some improvements on that, I'm closing this PR.

PR #52, opened 2014-11-06
Update Mojos to resolve artifacts via Aether

achim86:
According to my Question on the [Mailing List](http://karaf.922171.n3.nabble.com/Question-about-artifact-resolving-in-karaf-maven-plugin-td4036242.html) I updated the remaining goals to use Aether resolution.

I have two remarks on the changes. 

1. In general there were two ways how artifacts could be resolved. Either via the provided project repositories (default case) or via a repository which is dedicated to the artifact (if not null). The currently provided interface for resolving artifacts via Aether (`org.apache.karaf.tooling.features.DependencyHelper`) does not provide any way to edit the repositories which are used to resolve artifacts. The resolver always takes the project repositories which are defined as read only. First I thought the right way would be to extend the interface to take a List of repositories to resolve artifacts against. This would allow me to pass the artifact's repository to the resolver. When trying to test my changes I did not find any use case when such an artifact repository is used instead of the project repositories. In the [git history](http://svn.apache.org/viewvc?view=revision&revision=826783) I found that this functionality was introduced in 2009 when the Plugin was part of Felix. Bundle URI must have looked like `<repo>!<dependency>`. I am not sure whether this is relevant any longer. I first made the decision to take back the changes from the interface and remove the functionality to resolve artifacts using their dedicated repository.
2. I insufficiently tested Goal [features-validate-descriptor](http://karaf.apache.org/manual/latest/developers-guide/karaf-maven-plugin-features-validate-descriptor.html) as linked features can not be resolved. This is a major problem as Karaf Standard and Enterprise features are added to the validation by default (the same applies for 4.0.0.M1).

### DIFF
--- a/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/features/AbstractFeatureMojo.java
+++ b/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/features/AbstractFeatureMojo.java
@@ -17,9 +17,9 @@
  */
 package org.apache.karaf.tooling.features;
 
+import java.io.File;
 import java.net.URI;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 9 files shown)


jbonofre:
It's the case on K3 and K4. I close this PR.

PR #55, opened 2015-02-17
Cloudyle 3.0.x

agrz:
I added the function to karaf maven plugin to allow replacing dependency versions in feature.xml.
This is useful to add configuration files from maven artifacts. For each dependency in the pom.xml a property is created automatically during plugin execution. The properties are named artifactId.version (or groupId.artifactId.version).
This allows a syntax in feature.xml like this:
<configfile finalname="etc/test.cfg" override="false">mvn:test/dependency-module-c/${test.dependency-module-c.version}/cfg</configfile>


jbonofre:
Do you mind to create a PR based on master ?

I'm closing this PR as karaf 3.0.x is in maintenance mode now.

PR #56, opened 2015-03-05
KARAF-3590: Don't log Passwords in clear text

chirino:
* Bundles can now register CommandLoggingFilter objects in the OSGi registry to get a chance to filter out sensitve data data from logs before the logging happens.


jbonofre:
It has been fixed: can you close this PR ? Thanks.

PR #72, opened 2015-07-26
Karaf-3882: pem for ssh host keys

bimargulies:
This works.


jbonofre:
Review in progress ...

PR #74, opened 2015-08-04
feature:export-bundles

bdgould:
Added a command to the feature subset that will allow a user to export all of the bundles associated with a given feature to a specified location on their file system.  The use case that inspired this functionality involved the creation of an OBR repository (we used Cave) that hosted the bundles necessary to run an camel workflow making use of an activemq JMS broker.


jbonofre:
Review in progress. I added legal header (apache license) in the class.

PR #80, opened 2015-09-09
KARAF-3982: Add support for editing property files to the assembly mojo.

bimargulies:
This appears to be all done.


maggu2810:
Hello, what do you think about the append / prepend hint?

bimargulies:
Done

tooling/utils/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+        See the License for the specific language governing permissions and
+        limitations under the License.
+    -->
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.apache.karaf.tooling</groupId>
+        <artifactId>tooling</artifactId>
+        <version>4.0.2-SNAPSHOT</version>
```
maggu2810: This should be changed to match the project version of the current master.

maggu2810:
Hi @bimargulies,
what about variable substitution? Using e.g. ${project.version} e.g. in an added value for a property seems a valid and perhaps common use case. What do you think about?

maggu2810:
e.g. a attribute for the value tag "substitue" to switch the substitution (InterpolationHelper.substVars(...)) on / off.

PR #91, opened 2015-11-09
Fixes KARAF-4091: Support restarting the Karaf JVM and updating it's …

chirino:
…lib directory

The karaf.restart.jvm.supported=true system property will be set if restarting the JVM is supported.  If karaf is shutdown and the karaf.restart.jvm=true system property is set, then karaf will exit with error code = 10.  The shell scripts will check for this error code and restart Karaf.

On unix, sometimes it's desirable to give up this feature so that the JVM process can be 'exec'ed by the shell script for simpler process management (For example if your running in a Docker container).  In these cases you can set the KARAF_EXEC=exec env var to return to the current behavior of using exec to launch the java process.

### DIFF
--- a/assemblies/features/base/src/main/resources/resources/bin/karaf
+++ b/assemblies/features/base/src/main/resources/resources/bin/karaf
@@ -417,7 +417,42 @@ run() {
     fi
     cd "$KARAF_BASE"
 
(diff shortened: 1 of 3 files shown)


jbonofre:
It has been fixed by this commit: https://git-wip-us.apache.org/repos/asf?p=karaf.git;a=commit;h=a47c36b7f00c6b593b7a9b5a018deda4aeca1cc0

I'm closing this PR.

PR #92, opened 2015-11-12
[KARAF-4107] Upgrade Json to version 20150729

oscerd:
Hi all,

This PR is related to:
https://issues.apache.org/jira/browse/KARAF-4107

Andrea


jbonofre:
In order to verify the feature, we have to use <javase>1.8</javase>.
As now master depends to Java8 (Karaf 4.1.x), I will do the change there.

PR #123, opened 2015-12-27
[KARAF-4199] Privacy Violation: Heap Inspection

oscerd:
Hi all,

This PR is related to:
https://issues.apache.org/jira/browse/KARAF-4199

Andrea


CodingFabian:
Hi Andrea, how did you verify this solves the issue?
in both cases answers[i] contains a reference to a string object containing the same string contents.

If you consider this a valid attack vector, then you should only use char[] and clear that one after the password has been verified.
your fix attempts this, but fails because the end result is still a string in memory.

(for more information: http://stackoverflow.com/questions/8881291/why-is-char-preferred-over-string-for-passwords-in-java)

oscerd:
You're right. I read you comment in JIRA too, so I close this PR as unuseful.

PR #126, opened 2016-01-01
[KARAF-2971] there is no MBean operation for easily get bundle status

oscerd:
Hi all,

This PR is related to:
https://issues.apache.org/jira/browse/KARAF-2971

Andrea


jbonofre:
I changed this PR:
1. the MBean interface has to be changed (not only the implementation)
2. I'm changed the BundleService even if we already have
3. No itest at least on the command
I fixed that and the push will close this PR.

oscerd:
I see the method also in BundleServiceMBean but maybe I lost something.

I'm waiting for your commits to see where I was wrong and do a better work
next time.

Thank you
Il 13/gen/2016 07:45, "Jean-Baptiste Onofré" <notifications@github.com> ha
scritto:

jbonofre:
My bad, you are right, I missed it. I added the itest (build in progress).

PR #127, opened 2016-01-09
Fixed typo in javadoc

Brimstedt:


jbonofre:
The Javadoc looks good to me: @Reference matches services in the service registry.

Brimstedt:
Not a big deal, but I dont get it :-)

@Service annotation has @Target(value={TYPE}) so it cant be an annotation
of a field?
(As stated in "A class annotated with @Service can have fields annotated
with @Service ")

PR #128, opened 2016-01-10
[KARAF-4255] Only include conditional dependencies in assembly if condition is met

CodingFabian:


jbonofre:
Fair enough to consider only the boot features.

PR #129, opened 2016-01-15
KARAF-4270: Don't NPE for commands without descriptions.

bimargulies:


maggu2810:
I found this PR after I implemented the fix myself.
The other branches in that file (that handles other command types) implement the getDescription() that it returns getName() if the description is null.
Shouldn't we do here the same to handle it the same way as the other ones?

PR #135, opened 2016-01-21
[KARAF-129] Add upgrade option to Feature install command

vladimirfx:
Feature upgrade implemetation via install command option.

### DIFF
--- a/features/command/src/main/java/org/apache/karaf/features/command/InstallFeatureCommand.java
+++ b/features/command/src/main/java/org/apache/karaf/features/command/InstallFeatureCommand.java
@@ -52,6 +52,9 @@ public class InstallFeatureCommand extends FeaturesCommandSupport {
     @Option(name = "-t", aliases = "--simulate", description = "Perform a simulation only", required = false, multiValued = false)
     boolean simulate;
 
+    @Option(name = "-u", aliases = "--upgrade", description = "Perform an upgrade of feature if previous version are installed or install it", required = false, multiValued = false)
+    boolean upgrade;
+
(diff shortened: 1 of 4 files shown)


jbonofre:
Karaf 4.x already likely supports feature upgrade (thanks to the feature lifecycle).

vladimirfx:
May be it is not so clear stated in documentation. How I can upgrade feature from one version to another without uninstall/install it entirely?
Main idea behind this patch is minimizing changes of current bundle cache state. We are changing feature version in requirements and put it to deployer. Deployer calculate minimum required changes and execute it. No useless refreshes, un/installs etc. is performed.

vladimirfx:
Currentl documentation states:
> You can update a release by installing the same feature (with the same SNAPSHOT version or a different version).

> Thanks to the features lifecycle, you can control the status of the feature (started, stopped, etc).

> You can also use a simulation to see what the update will do.

But what hepend with old feature when new is installed? It will stay here and should be manualy uninstalled. But old feature may contains singleton bundles or something like plugins with previous version or fragments that may be incorrectly associated with the new feature bundles. On my experience there is many features that not support two versions installed at the same time.
By un/installing, starting/refreshing bundles in one deployer "transaction" we achieve behaviour closest to clean uninstall/install cylce without useless install/unistall stop/start operations on unchanged bundles.
Actualy this patch writen for Karaf 4.0.0 and successfuly used by our ops team in production. Main use case is upgrade of customer own features with business functionality that contains many relatively "static" dependencies: unistall 130+ bundles and install it again is not so optimal... Installation of new version and then delete old version is not possible because there 1 singleton bundle (e.g. DataNucleus Core) that can not be installed in parallel.

PR #138, opened 2016-01-26
KARAF-3798 - Use a single threaded executor for FeaturesService

CodingFabian:
prevents potential concurrent installations, each provision is queued on
a single executor thread.

### DIFF
--- a/features/core/src/main/java/org/apache/karaf/features/internal/osgi/Activator.java
+++ b/features/core/src/main/java/org/apache/karaf/features/internal/osgi/Activator.java
@@ -273,6 +273,7 @@ public class Activator extends BaseActivator {
         }
         super.doStop();
         if (featuresService != null) {
+            featuresService.stop();
             featuresService = null;
         }
         if (digraph != null) {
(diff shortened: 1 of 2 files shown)


jbonofre:
R: @jbonofre

CodingFabian:
As said in the Jira, another option would be to still use a cached executor pool, but still avoid creating large thread pools over and over again:
screen shot 2016-06-24 at 12 57 40
(those thread pools have been created by the FeaturesService, not visible in this view but in thread tracking of jprofiler)

jbonofre:
I would prefer a cache executor pool. However, I'm afraid about the number of threads created. Let me experiment a bit.

CodingFabian:
When I start up my karaf (using a certains et of features/bundles) the provision calls spawn 77 threadpools * coresize (8 here at the moment) = 616 threads.
A customer with a 256 core machine observed a massive native memory usage (probably because a thread uses 1mb of stack and there were 20k of them)

CodingFabian:
IMHO the situation only gets better by moving the pool creation from the doProvisionInThread() to activation

jbonofre:
Fair enough.

jbonofre:
LGTM

PR #144, opened 2016-02-12
[KARAF-4330] Instance script doesn't return correct PID number if roo…

oscerd:
…t instance is started two times

I submit this little PR for review. In case we start root instance two times, the instance script trace the second instance PID even if the process is hanging because of the first instance.

If it's ok for everyone I'll push this on master, 4.0.x and perhaps on 3.0.x.

I'm waiting for your feedback :-) Thanks.

Andrea


oscerd:
The errors in Jenkins build seems to be sporadic. I guess the PR is ok.

PR #156, opened 2016-02-24
Fix presentation names

skahmann:
OBR R5 resources do not have Presentation Names so various OBR commands throw exceptions or present null to user.


skahmann:
Fix for https://issues.apache.org/jira/browse/KARAF-4357

skahmann:
Fixed.

PR #157, opened 2016-02-25
Add notice about wrap feature

yodamad:
Add a notice to inform that wrap feature is a prerequisite to manage bundle installed using wrap protocol


splatch:
I've pushed this commit to master with changed commit message cause it did not contain issue id. Feel free to close it if robot won't do that.

PR #160, opened 2016-03-02
KARAF-4374, Fix problems in Karaf start script

lkiesow:
This commit fixes most of the problems pointed out by shellcheck [1] in
the karaf start script (bin/karaf). These include:

 - Properly quote variables to prevent spaces from breaking the script
 - Ensure that KARAF_HOME is set for commands which would otherwise try
   to delete parts of the root file system. E.g: `rm -rf "$KARAF_HOME/lib"`
 - Use only $(..) instead of both $(...) and \`...\`
 - Require bash since not all commands are posix compliant anyway.

[1] http://shellcheck.net

### DIFF
--- a/assemblies/features/base/src/main/resources/resources/bin/karaf
+++ b/assemblies/features/base/src/main/resources/resources/bin/karaf
@@ -16,26 +16,26 @@
 #    limitations under the License.
 #
 
-DIRNAME=$(dirname "$0")
-PROGNAME=$(basename "$0")
+DIRNAME=$(dirname "${0}")
+PROGNAME=$(basename "${0}")
 
 #
 # Sourcing environment settings for karaf similar to tomcats setenv
 #
 KARAF_SCRIPT="karaf"
 export KARAF_SCRIPT
-if [ -f "$DIRNAME/setenv" ]; then
-  . "$DIRNAME/setenv"
+if [ -f "${DIRNAME}/setenv" ]; then
+  . "${DIRNAME}/setenv"
 fi
 
 #
 # Set up some easily accessible MIN/MAX params for JVM mem usage
 #
-if [ "x$JAVA_MIN_MEM" = "x" ]; then
+if [ "x${JAVA_MIN_MEM}" = "x" ]; then
     JAVA_MIN_MEM=128M
     export JAVA_MIN_MEM
 fi
-if [ "x$JAVA_MAX_MEM" = "x" ]; then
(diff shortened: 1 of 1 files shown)


maggu2810:
> * Use only $(..) instead of both $(...) and `...`

What do you think about:
* Use only ${VAR} instead of both ${VAR} and $VAR

lkiesow:
@maggu2810, I do not have any strong feelings about that but since it is only one `sed` command away, I created a patch for that as well :)

maggu2810:
Cool :wink: ($0 could be also replaced by ${0}).

I hope that the other ones like this too.
I do not have a strong feeling about that (`...` and $(...); $VAR and ${VAR}), too.
But I think it is good to be consistent.

lkiesow:
Fixed that as well.

The ticks are not that bad, I guess. You can find a lot of advice to abandon them in favor of `$(...)` but technically, they do not introduce any problems.

I decided to submit a patch only after I realized that this script could possibly try to delete parts of my root file system. That should never happen under normal circumstances, but better be safe than sorry. And when I was at it, I thought, I might as well change the ticks.

jbonofre:
Changing /bin/sh to /bin/bash means that we can break support one some system (for instance Solaris by default). Is it really required ?

lkiesow:
Hi @jbonofre,
I did that on purpose since the script is already using some non-posix compliant functions which bash provides. Namely `ulimit -H`, `ulimit -n` and type`. I think it makes more sense to have the script fail at the beginning rather than having a possibly undefined behavior later

I do not now how important Posix compliance is for Karaf and if we want to try fixing those problems instead.

jbonofre:
It makes sense. Less than Posix, the purpose is that Karaf scripts work on most Unix system as possible (now they work on Linux, MacOS, Solaris, AIX, *BSD). Let me review your proposal and see how it works on different VM.
Thanks.

jbonofre:
Can you please rebase your PR as it has conflicts with the repo ? Thanks !

lkiesow:
Sure, will do that as soon as I am home again tomorrow evening. 
Lars

jbonofre:
No rush, whenever you have time. Thanks !

lkiesow:
Rebased the branch.
Should now merge without problems.

lkiesow:
@jbonofre, is there anything still missing?

jbonofre:
Sorry, let me double check to merge. Thanks !

PR #164, opened 2016-03-16
[KARAF-4423] Jaas AutoEncryption support can correctly shutdown and i…

CodingFabian:
…s only started when needed.

### DIFF
--- a/jaas/modules/src/main/java/org/apache/karaf/jaas/modules/impl/Activator.java
+++ b/jaas/modules/src/main/java/org/apache/karaf/jaas/modules/impl/Activator.java
@@ -71,13 +71,16 @@ public class Activator extends BaseActivator implements ManagedService {
 
         karafRealm = new KarafRealm(bundleContext, config);
         register(JaasRealm.class, karafRealm);
-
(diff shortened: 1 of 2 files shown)


jaas/modules/src/main/java/org/apache/karaf/jaas/modules/impl/Activator.java
```diff
@@ -71,13 +71,14 @@ protected void doOpen() throws Exception {
...
-        autoEncryptionSupport = new AutoEncryptionSupport(config);
+        if (Boolean.parseBoolean((String) config.get(ENCRYPTION_ENABLED))) {
+          autoEncryptionSupport = new AutoEncryptionSupport(config);
+        }
     }
 
     @Override
     protected void doStop() {
-        StreamUtils.close(autoEncryptionSupport);
+        autoEncryptionSupport.close();
```
akuhtz: Check for null missing?

CodingFabian: right!

jbonofre:
LGTM

PR #168, opened 2016-03-26
KARAF-4447

mhautman:


jbonofre:
BUILDING is specific to one Karaf version, so, it should be updated (not since Karaf 4).

PR #169, opened 2016-03-26
KARAF-4444 - Make a new command service:get

mhautman:


cschneider:
What is the purpose of this command?

It looks like the command simply lists services that match a name which is the same that service:list does. There is also already a command service:get to retrieve a service.

mhautman:
Indeed Christian.

The original problem was that when trying to know how to use service:get with "service:get --help" it threw an error.

IMHO , scr:list should print every service and not a specific one. That's what the user may think also, that's why I created a service:get (which already existed but now it's a nice command..)

cschneider:
I think service:list works fine. By default it prints all serivices and you can also set a filter. It does not make sense to replicate this in service:get. The purpose of service:get is to get the service object which then allows to call the service on the console. We need to improve the help and some examples.

jbonofre:
Why a new service:get command ? It already exists: it's an alias in etc/shell.init.script:

service:get = { $.context getService ($.context getServiceReference $args) };

IMHO, the purpose of the Jira is to fix the Console Impl in case of alias (for instance, the help could be "This command is an alias to ...").

mhautman:
I'm sorry guys.. I understood it wrong.
I will provide another PR with the correct changes.

jbonofre:
There's no problem ! We are here to help you and review !

PR #184, opened 2016-05-11
Karaf-4524: SCTP class from JDK missing in jre.properties

frouleau:
JDK is using com.sun.nio.sctp package for sctp support which is not in jre.properties.


mhautman:
Hi @frouleau,

FYI, you branch doesn't seem to be synced with the Apache master branch.

Regards,
Morgan

frouleau:
Yep I have made the commit against 4.0.x branch by mistake. I am fixing it

PR #185, opened 2016-05-11
[KARAF-4524] Add missing jdk's sctp classes to jre.properties

frouleau:
JDK is using com.sun.nio.sctp package for sctp support which is not in jre.properties.


jbonofre:
Does this package exist in other JDK (like IBM or OpenJDK) ? I gonna check.

frouleau:
It exists in openJdk but for IBM I don't know.

jbonofre:
It seems like the package exists on IBM JDK (at least on AIX).

PR #188, opened 2016-05-19
KARAF-4442 - improving misleading message when updating a SNAPSHOT

mfrazier63:
Improved misleading message per JIRA KARAF-4442


jbonofre:
R: @jbonofre

PR #195, opened 2016-06-03
[KARAF-884] Generate feature dependencies

skitt:
Currently, POMs containing feature dependencies (classifier "features"
and type "xml") produce features listing all the required bundles.
This patch causes feature generation to represent feature dependencies
using <feature/> elements instead.

As a side-effect, aggregate features produce a new aggregate feature
containing all the generated features.


jbonofre:
R: @jbonofre

PR #196, opened 2016-06-08
KARAF-4564 Can't start karaf using symbolic link

adetalhouet:
When executing the karaf script, it gets the DIRNAME based on $0 which is the path used to start the script. This DIRNAME is then used to set the KARAF_HOME and multiple other KARAF_* env variables. 
Using a symbolic link, you would have, for instance, usr/bin/karaf redirecting to /opt/opendaylight/bin/karaf. So $0 would be usr/bin and not /opt/opendaylight/bin so the locateHome function isn't setting the right path for the KARAF_HOME. 

This ends up failing to start ODL with following 
ERROR: Error: Could not find or load main class org.apache.karaf.main.Main

Signed-off-by: Alexis de Talhouët <adetalhouet@inocybe.com>

### DIFF
--- a/assemblies/features/framework/src/main/filtered-resources/resources/bin/client
+++ b/assemblies/features/framework/src/main/filtered-resources/resources/bin/client
@@ -16,8 +16,9 @@
 #    limitations under the License.
 #
 
-DIRNAME=`dirname "$0"`
-PROGNAME=`basename "$0"`
+REALNAME=`realpath -e "$0"`
+DIRNAME=`dirname "$REALNAME"`
+PROGNAME=`basename "$REALNAME"`
 
 #
 # Sourcing environment settings for karaf similar to tomcats setenv
--- a/assemblies/features/framework/src/main/filtered-resources/resources/bin/client
+++ b/assemblies/features/framework/src/main/filtered-resources/resources/bin/client
@@ -16,7 +16,7 @@
 #    limitations under the License.
 #
 
-REALNAME=`realpath -e "$0"`
+REALNAME=`readlink -e "$0"`
 DIRNAME=`dirname "$REALNAME"`
 PROGNAME=`basename "$REALNAME"`
 
--- a/assemblies/features/framework/src/main/filtered-resources/resources/bin/instance
+++ b/assemblies/features/framework/src/main/filtered-resources/resources/bin/instance
@@ -16,8 +16,9 @@
 #    limitations under the License.
 #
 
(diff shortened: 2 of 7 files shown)


adetalhouet:
The pull request isn't complete as it should take care of the .bat scripts a well. But I wasn't able to figure it out for the batch script.

jbonofre:
Unfortunately, I don't think it's a good idea to use `realpath` as it's not available on most of Unix systems.
For instance, on Debian/Ubuntu, it requires:

```
$ sudo apt-get install realpath
```

Maybe `readlink` is more appropriate ?

jbonofre:
Even better, we can test if `readlink` is available on the system, then use it, else fallback to the current behavior.

adetalhouet:
Sure, would you like me to test it to make sure it solves the issue the same way?

jbonofre:
That would be great ! Thanks !

adetalhouet:
I can confirm the `readlink` is working fine with the three environment mentioned in the JIRA

jbonofre:
As said in my previous comment, even if `readlink` is available by default on most of system, it's not necessarily the case. I would add a test to check if `readlink` command is available: if yes, then we use it for `basedir`, else we "fallback" to the current behavior.

For instance, something like `type readlink` and check if it returns 0 (`readlink` is found on the system) or 1 (`readlink` is not found on the system).

adetalhouet:
So would this be more appropriate?
`
READLINK=`type readlink`
if [ "x$READLINK" = "x" ]; then
    REALNAME=$0
else
    REALNAME=`readlink "$0"`
fi

DIRNAME=`dirname "$REALNAME"`
PROGNAME=`basename "$REALNAME"`
`

jbonofre:
I would do more something like:

```
REALNAME=`readlink "$0"`
if [ $? != 0 ]; then
  REALNAME=$0
fi
```

jbonofre:
I updated the PR with:

```
REALNAME=`readlink -e "$0" > /dev/null 2>&1`
if [ $? != 0 ]; then
    REALNAME=$0
fi
DIRNAME=`dirname "$REALNAME"`
PROGNAME=`basename "$REALNAME"`
```

It works fine. Rebasing and merging.

adetalhouet:
@jbonofre Thank you for such a great support!

jbonofre:
@adetalhouet you are welcome ! My pleasure !

vorburger:
=> https://issues.apache.org/jira/browse/KARAF-4599

PR #197, opened 2016-06-08
KARAF-4566 "karaf" script invokes /bin/sh but requires /bin/bash functions

adetalhouet:
The bin/karaf script uses the "local" command which is a shell builtin of bash and similar shells, but is not required for POSIX-compliance in sh. When I attempt to run karaf on a Solaris system, I see the following output:

root@solaris:/opendaylight/bin# ./karaf
./karaf[172]: local: not found [No such file or directory]
./karaf[182]: local: not found [No such file or directory]
./karaf[183]: local: not found [No such file or directory]

Lines 172, 182 and 183 invoke "local" to make local variables to the function. According to "man bash", this is a shell builtin. However, bin/karaf is invoked as:

#!/bin/sh

On most flavors of linux, this resolves to bash or dash which probably runs in a restricted environment after checking to see that its $0 is sh. But on Solaris's /bin/sh is actually ksh93 for backwards compatibility.

Since "local" is not part of a POSIX-compliant /bin/sh, depending on it in a script that is invoked with /bin/sh is a bug.

(this explaination is borrowed from https://issues.apache.org/jira/browse/MNG-5852)

Signed-off-by: Alexis de Talhouët <adetalhouet@inocybe.com>

### DIFF
--- a/assemblies/features/framework/src/main/filtered-resources/resources/bin/client
+++ b/assemblies/features/framework/src/main/filtered-resources/resources/bin/client
@@ -162,7 +162,7 @@ setupNativePath() {
 }
 
 pathCanonical() {
-    local dst="${1}"
+    dst="${1}"
     while [ -h "${dst}" ] ; do
         ls=`ls -ld "${dst}"`
         link=`expr "$ls" : '.*-> \(.*\)$'`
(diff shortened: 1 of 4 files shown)


adetalhouet:
@jbonofre Thank you for such a quick fix :)

PR #198, opened 2016-06-13
KARAF-4569: OSGi framework is not shut-down gracefully on SIGTERM

lburgazzoli:


CodingFabian:
wouldn't a shutdown hook accomplish the same?

lburgazzoli:
Shutdown Hook was my first attempt, but it turned out not to be the same maybe because of the ordering (i.e. the framework may have installed one on its own)

jbonofre:
R: @jbonofre

lburgazzoli:
@jbonofre any finding I should fix ?

jbonofre:
Merging

PR #200, opened 2016-06-14
KARAF-4571 Use version range in feature.xml if useVersionRange is true

SourcePond:
This pull-request implements KARAF-4571. It would be cool if somebody could review it.


SourcePond:
Build could not be executed because no build processor could be acquired. What should I do?

jbonofre:
R: @jbonofre

PR #204, opened 2016-06-27
karaf-service should return 0 on start when service is already running

dhx:
tools like puppet have problems with a service not returning 0 on start when it is already running.


jbonofre:
LGTM

jbonofre:
Are you sure about this change ?

0 means that the process finished correctly. In the case of the service already running, we return 1 to notify the user that something is not correct.

dhx:
exactly, see also in the description of the commit:
```
from: http://refspecs.linuxbase.org/LSB_3.1.1/LSB-Core-generic/LSB-Core-generic/iniscrptact.html
... In addition to straightforward success, the following situations are also to be considered successful:
...
- running start on a service already running
...
```

puppet for example relies on this behavior, and its consistent whith all the other init scripts I've seen so far.

jbonofre:
OK, fair enough.

LGTM

PR #205, opened 2016-06-30
[KARAF-4600] RBAC - MBean fails to resolve ACL if the order of properties in object name differs

tadayosi:
https://issues.apache.org/jira/browse/KARAF-4600

This fix makes `KarafMBeanServerGuard` always use the canonical object name when resolving ACL.


tadayosi:
See also: https://issues.jboss.org/browse/ENTESB-5775

PR #214, opened 2016-07-25
Fix karaf-maven-plugin attaching the same artifact multiple times

CMoH:
The problem occurs when creating assemblies with karaf-maven-plugin version 4.0.4 (and I assume later on by looking at the code). One can examine the output of mvn -X and look for the attachedArtifacts list to confirm that assembly artifacts appears twice.

The effect is that maven-install-plugin installs each artifact (zip and tar.gz, for instance) twice in the local repository, which does not cause any visible problem. It also seems to check if the artifacts are the same and skip duplicates.

However, maven-deploy-plugin fails when attempting to upload a release version twice (if the repository server is configured to reject release overwrites), which in turn failed my run of maven-release-plugin.

I hope the above explains the changes I submitted to you, and that you will consider applying them.


jbonofre:
R: @jbonofre

PR #225, opened 2016-08-08
KARAF-4649: AssemblyMojo : blacklistPolicy set to null if not defined in pom

lburgazzoli:


jbonofre:
R: @jbonofre

lburgazzoli:
@jbonofre thx

PR #228, opened 2016-08-22
[KARAF-1583] Add repository dependencies

skitt:
Any repository dependencies encountered while processing feature
dependencies are added to the repositories. We track
feature-repository correspondences and only add necessary repositories
after processing all the features, so we end up adding repositories
necessary to find feature dependencies, but nothing else.

Repositories are tracked explicitly to allow later integration of
repositories mentioned in pre-existing feature.xml files; this will
allow large feature repositories to be used optimally.


jbonofre:
R: @jbonofre

PR #229, opened 2016-08-23
Fix typo in wrapper install output message

glefloch:
Fix the `rm` command to use the karaf-service file instead of the complete path


jbonofre:
R: @jbonofre

PR #245, opened 2016-09-29
MavenEntries will not be transferred

rekx:
Hi!

It seems that the `MavenArchiveConfiguration` which i configured in my pom.xml will not be transferred in the final `MANIFEST.MF`. This should fix it.


jbonofre:
R: @jbonofre

PR #246, opened 2016-10-03
[KARAF-4748] Make Felix Resolver Threads configurable and reuse Executor

CodingFabian:
By setting `resolverThreads` in `org.apache.karaf.features.cfg` one can
influence the amount of threads that Felix resolver will use.
This change also re-uses the same ExecutorService instead of creating a new
one with every resolve() call.


CodingFabian:
closing this PR. Will work first on resolving this in felix.

PR #248, opened 2016-10-12
[KARAF-4764] Add ability to configure RMI Registry locate and create …

netudima:
…flags


jbonofre:
R: @jbonofre

jbonofre:
The default `etc/org.apache.karaf.management.cfg` should at least contain comments with the properties (kind of documentation). I do this change.

netudima:
Got it, thank you.

PR #251, opened 2016-10-20
[KARAF-4773] AbstractRetryableDownloadTask is smarter with pax-url-ae…

grgrzybek:
…ther 2.5.0


grgrzybek:
This is fixed in `master`: e4f9ead

PR #254, opened 2016-10-24
[KARAF-4487] - Gssapi

acartapanis:
A GSSAPI login module.
Only tested with Kerberos authentication.
Typically allows to use an Active Directory Server with Kerberos Authentication as LDAP backend.
https://issues.apache.org/jira/browse/KARAF-4487


acartapanis:
The tests failure is not reproductible locally. 
Probably coming from a host mismatch...

jbonofre:
R: @jbonofre

acartapanis:
Test failure seems solved

PR #255, opened 2016-10-26
[KARAF-4805] configfiles are not copied to system directory

tterm:
Added the missing copy of the config files and tested it.


jbonofre:
R: @jbonofre

jbonofre:
For the record, it's better to do a PR based on master (and not another branch) as it's used by the Apache GitHub integration.

PR #256, opened 2016-10-27
KARAF-4806: specify /bin/bash for scripts with bashisms

skitt:
The client, instance and shell scripts use ulimit and type, but these
are bashisms. In some environments (e.g. Solaris, Debian) /bin/sh
isn't bash.

Signed-off-by: Stephen Kitt skitt@redhat.com

### DIFF
--- a/assemblies/features/base/src/main/filtered-resources/resources/bin/client
+++ b/assemblies/features/base/src/main/filtered-resources/resources/bin/client
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 #    Licensed to the Apache Software Foundation (ASF) under one or more
 #    contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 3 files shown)


assemblies/features/base/src/main/filtered-resources/resources/bin/client
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
geomacy: Perhaps better `#!/usr/bin/env bash`?

geomacy: (although bash may not be on all platforms, eg. restricted ones for containers, perhaps best just get rid of the bashisms!)

jbonofre: I would prefer:
1. directly refer `/bin/bash`
2. remove bash dependency.
It's what I tried at the beginning but due to some limitation and improvements, we introduced bash. I gonna check what's bash specific and fix that.

jbonofre:
-1 on this PR: {{/bin/bash}} is not available on FreeBSD for instance. I would prefer to stay with {{/bin/sh}} and remove all bash dependencies in the scripts.

PR #257, opened 2016-10-27
KARAF-4806: specify /bin/bash for scripts with bashisms

skitt:
The client, instance and shell scripts use ulimit and type, but these
are bashisms. In some environments (e.g. Solaris, Debian) /bin/sh
isn't bash.

Signed-off-by: Stephen Kitt skitt@redhat.com
(cherry picked from commit 9e72b9357d20de866abce7391cdc41ba616f3a75)

### DIFF
--- a/assemblies/features/base/src/main/filtered-resources/resources/bin/client
+++ b/assemblies/features/base/src/main/filtered-resources/resources/bin/client
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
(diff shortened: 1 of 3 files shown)


oscerd:
Usually the commit on master is backported on the older branches. So maybe this PRs on karaf-4.0.x and 3.0.x can be avoided.

jbonofre:
Can you please close this PR ? Only a PR on master is required (github integration only consider master). Thanks !

PR #259, opened 2016-10-28
KARAF-4809, SSH should not listen to all hosts

lkiesow:
The default SSH server configuration will make Karaf listen to all
hosts. It is usually good practice to instead listen to localhost only
by default to avoid possible security risks (e.g. accidentally exposing
an unconfigured SSH server).


jbonofre:
Can you please close this PR ? Only a PR on master is required. Thanks.

lkiesow:
Sure. It's closed.

PR #262, opened 2016-11-02
KARAF-4813, RMI should not listen to all hosts

lkiesow:
The default RMI configuration will make Karaf listen to all hosts.
Instead. it is usually good practice to, by default, listen to localhost
only to avoid possible security risks.

This patch modified the default configuration file to set RMI to
localhost.


jbonofre:
R: @jbonofre

PR #268, opened 2016-11-24
[KARAF-4843] Updating factory configuration leads to new configurati…

cfiehe:
…on instance


cfiehe:
This fix requires "[KARAF-4846] Property karaf.etc must be a canoncial path", otherwise Fileinstall cannot find the corresponding configuration file and a duplicate configuration gets created.

cfiehe:
This fix adds the new argument `--alias` to the config shell command, which allows to set the alias for a factory configuration used by Felix Fileinstall to construct the backing file in the config folder. The filename follows the scheme `<factoryPid-alias>.cfg`. When no alias is provided a synthetic one without any hyphens gets created. The solution relies on the fact that Fileinstall uses the property `felix.fileinstall.filename` to map from the configuration file to the actual configuration and vice versa.

In my use case, I make heavy use of factory configurations that are getting updated multiple times. The current implementation has the drawback that the filename created by Felix Fileinstall contains the actual `pid` of the configuration instance. The `pid` comprises a UUID, that uses hyphens as separator leading to the problem that some parts of the `pid` get interpreted as `factoryPid`.

PR #270, opened 2016-11-26
KARAF-4853, Option to prevent execution as root

lkiesow:
Usually, it is not a good idea to run Karaf as root. In some cases, it
might even be harmful. This patch introduces the option `KARAF_NOROOT`
which, if set, will prevent Karaf from being executed as root.


jbonofre:
R: @jbonofre

gnodet:
I don't really understand the patch.

lkiesow:
@gnodet, I'm sorry. It was my mistake. You are right, it did make no sense. It's fixed now.

jbonofre:
@lkiesow thanks. I'm waiting for the update.

lkiesow:
@jbonofre, the pull request is already updated.

jbonofre:
Thanks, reviewing and checking the Jenkins failure.

PR #274, opened 2016-12-19
KARAF-4897 - Add an option to simplify bundle dependencies

skitt:
This adds a simplifyBundleDependencies option to
GenerateDescriptorMojo, disabled by default, which drops bundle
dependencies from features if the bundles are included in "parent"
features. I.e. if feature A depends on bundle B and feature C, and
feature C also includes bundle B, then feature A will be reduced to
only depend on feature C, relying on the latter to supply bundle B.


skitt:
If this is acceptable for master, it would be great to have it in Karaf 4.0 too — https://github.com/skitt/karaf/tree/remove-duplicate-bundles has the appropriate changes.

jbonofre:
R: @jbonofre

jbonofre:
Completing the review.

jbonofre:
I just tested the PR and it looks good. The Jenkins failure is not related to the changes in the PR.
I'm merging and I will investigate the Jenkins issue.

PR #275, opened 2017-01-03
KARAF-4917 Support dynamicaliy registered configurations in RBAC

zregvart:
When adding configuration at runtime through
`ConfigurationAdmin#createFactoryConfiguration` Apache Felix appends
random UUID to the Configuration PID [1][2].

In order for those to be considered by `getRequiredRoles` that random 
UUID needs to be removed in comparison performed by `getGeneralPid`.

[1] https://github.com/apache/felix/blob/trunk/configadmin/src/main/java/org/apache/felix/cm/impl/ConfigurationManager.java#L397-L400
[2] https://github.com/apache/felix/blob/trunk/configadmin/src/main/java/org/apache/felix/cm/impl/ConfigurationManager.java#L1093-L1137


jbonofre:
R: @jbonofre

zregvart:
As per discussion in [KARAF-4917|https://issues.apache.org/jira/browse/KARAF-4917] one should use `getConfiguration` instead of `createFactoryConfiguration` to dynamically configure RBAC.

PR #276, opened 2017-01-31
[KARAF-4968] LDAPLoginModule does not correctly implement login method

sstrickx:
Correcting the behavior of the LDAPLoginModule's login method in case of failed authentication.
More details provided in the JIRA issue: https://issues.apache.org/jira/browse/KARAF-4968


jbonofre:
R: @jbonofre

jbonofre:
I'm  checking why Jenkins is failing.

PR #279, opened 2017-02-06
Add Aries Quiesce to aries-blueprint

rovarga:

### DIFF
--- a/assemblies/features/standard/src/main/feature/feature.xml
+++ b/assemblies/features/standard/src/main/feature/feature.xml
@@ -46,6 +46,7 @@
         <bundle start-level="20">mvn:org.apache.aries.blueprint/org.apache.aries.blueprint.cm/${aries.blueprint.cm.version}</bundle>
         <bundle start-level="20">mvn:org.apache.aries.blueprint/org.apache.aries.blueprint.core.compatibility/${aries.blueprint.core.compatibility.version}</bundle>
         <bundle start-level="20">mvn:org.apache.aries.blueprint/org.apache.aries.blueprint.core/${aries.blueprint.core.version}</bundle>
+        <bundle start-level="20">mvn:org.apache.aries.quiesce/org.apache.aries.quiesce.api/${aries.quiesce.api.version}</bundle>
         <conditional>
             <condition>bundle</condition>
             <bundle start-level="30">mvn:org.apache.karaf.bundle/org.apache.karaf.bundle.blueprintstate/${project.version}</bundle>
--- a/pom.xml
+++ b/pom.xml
@@ -227,6 +227,7 @@
         <aries.jndi.url.version>1.1.0</aries.jndi.url.version>
         <aries.proxy.version>1.0.5</aries.proxy.version>
         <aries.proxy.api.version>1.0.1</aries.proxy.api.version>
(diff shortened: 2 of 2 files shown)


rovarga:
@skitt I think we will need this one, or change it in our karaf packaging to resolve the problem of pax-exam tests running in parallel.

cschneider:
It shoud not be necessary. Did you find out how it makes a difference?

rovarga:
Essentially what we have is an blueprint extension which uses quiesce. When we try to test the feature containing it using pax-exam we see:

2017-02-06 00:25:29,005 | INFO  | pool-8-thread-1  | FeaturesServiceImpl              | 8 - org.apache.karaf.features.core - 4.0.7 | Refreshing bundles:
2017-02-06 00:25:29,005 | INFO  | pool-8-thread-1  | FeaturesServiceImpl              | 8 - org.apache.karaf.features.core - 4.0.7 |     org.apache.aries.blueprint.cm/1.0.8 (Wired to org.apache.aries.blueprint.core/1.6.2 which is being refreshed)
2017-02-06 00:25:29,005 | INFO  | pool-8-thread-1  | FeaturesServiceImpl              | 8 - org.apache.karaf.features.core - 4.0.7 |     org.apache.aries.blueprint.core/1.6.2 (Should be wired to: org.apache.aries.quiesce.api/1.0.0 (through [org.apache.aries.blueprint.core/1.6.2] osgi.wiring.package; filter:="(&(osgi.wiring.package=org.apache.aries.quiesce.manager)(version>=1.0.0)(!(version>=1.1.0)))"; resolution:=optional), org.eclipse.osgi/3.9.1.v20130814-1242 (through [org.apache.aries.blueprint.core/1.6.2] osgi.wiring.package; filter:="(osgi.wiring.package=org.osgi.service.framework)"; resolution:=optional))

This causes pretty much all of karaf to temporarily shut down, which confuses pax-exam into thinking the test case is done and proceeding with the next test case -- leading to races and large memory overhead, as we end up with multiple concurrent 'dead' karaf instances.

This may very well be the wrong approach, as it seems blueprint.core is installed via a different mechanism.

cschneider:
I can not imagine that it is necessary to have quiesce in the karaf blueprint feature to make this work.
Just add the quiesce api to to feature of your application under test. Karaf will then install quiesce together with blueprint and there should be no restart. Btw. almost no one is really using quiesce (besides IBM) and it might be removed in future versions of blueprint.

rovarga:
Well, the logs at https://logs.opendaylight.org/releng/jenkins092/controller-verify-carbon-mvn33-openjdk8/1779/archives/features/mdsal/features4-mdsal/target/surefire-reports/org.opendaylight.odlparent.featuretest.SingleFeatureTest-output.txt.gz clearly show that blueprint is being refreshed due to quiesce dependency becoming available. At any rate, if quiesce is on its way out, we'll just work this around in our packaging and get rid of its use when the time comes.

cschneider:
One other thing you can do quickly is to use the karaf-minimal distribution in the test. It does not include blueprint by default. So I think if you add a dependency to aries-blueprint and the quiesce bundle to your feature then a minimal distro should be able to start without refreshing the blueprint bundles.

rovarga:
Working with a custom distro fixed it up, thanks.

PR #282, opened 2017-02-14
Expose Dependency.hasVersion()

rovarga:
Since Dependency.getVersion() behaves the same as Feature.getVersion()
by returning "0.0.0" when no version is set, users attempting to examine
dependencies need to resort to touching the internal model to understand
that the version is not actually set.

Expose Dependency.hasVersion() to help potential users to understand
that the version has not been set.


jbonofre:
R: @jbonofre

jbonofre:
Can you create a PR based on master ?

PR #283, opened 2017-02-16
[KARAF-4989] Changed parsing of jaas ldap login module role.mapping option

valdar:
to support fqdn (i.e. strings with more than 1 equal char). The changes are compatible with previous logic.
An example usage is in: jaas/modules/src/test/java/org/apache/karaf/jaas/modules/ldap/LdapLoginModuleTest.java


valdar:
Thanks @mauiroma for the original code for the proposed fix.

jbonofre:
R: @jbonofre

PR #285, opened 2017-02-24
[KARAF-4997] Generator: merge dependency information

skitt:
In the generator, feature dependencies added from POM entries are
checked against existing dependency information (e.g. from
pre-existing feature.xml stubs). Matching entries are preserved and
the information merged: this allows versions to be determined from the
POM, and dependency/prerequisite flags to be defined in feature.xml.


jbonofre:
R: @jbonofre

PR #287, opened 2017-02-24
[KARAF-4999] Generator: ignore configfiles

skitt:
If a bundle matches a configfile, it shouldn't be processed as a
bundle. This patch ignores such bundles.

### DIFF
--- a/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/features/GenerateDescriptorMojo.java
+++ b/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/features/GenerateDescriptorMojo.java
@@ -47,6 +47,7 @@ import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.stream.XMLStreamException;
 
 import org.apache.karaf.features.internal.model.Bundle;
+import org.apache.karaf.features.internal.model.ConfigFile;
 import org.apache.karaf.features.internal.model.Dependency;
 import org.apache.karaf.features.internal.model.Feature;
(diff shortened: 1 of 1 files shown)


jbonofre:
R: @jbonofre

jbonofre:
Just to be sure, it means that the `features.xml` contains a bundle and a configfile with the same Maven coordinates ? Do you have an example ?

skitt:
Exactly. The issue is that the OpenDaylight build system (actually, our cross-project integration tests) needs to know the Maven artifacts which correspond to configfiles; so our POMs have entries like

        <dependency>
            
            <groupId>${project.groupId}</groupId>
            <artifactId>vbd-impl</artifactId>
            <version>${project.version}</version>
            <type>xml</type>
            <classifier>config</classifier>
        </dependency>

We also have a configfile entry in the feature.xml stub:

        <configfile filename="${configfile.directory}/vbd.xml">mvn:org.opendaylight.honeycomb.vbd/vbd-impl/${project.version}/xml/config</configfile>

This results in a feature.xml with the configfile and a wrapped bundle for the dependency:

        <configfile filename="etc/opendaylight/karaf/vbd.xml">mvn:org.opendaylight.honeycomb.vbd/vbd-impl/1.1.0-SNAPSHOT/xml/config</configfile>
        <bundle>wrap:mvn:org.opendaylight.honeycomb.vbd/vbd-impl/1.1.0-SNAPSHOT/xml/config</bundle>

This PR spots the latter and drops the bundle.

jbonofre:
Got it. Thanks !

skitt:
And all that means I really need to add IT tests for these ;-).

PR #288, opened 2017-02-28
Update AssemblyMojo.java

niclash:
HIbernate 5.x Features in the JBoss Maven repository https://repository.jboss.org/nexus/content/repositories/releases/ has a classifier of "karaf" and not "features".

This PR should fix that particular issue in the short term.

In case it is not obvious, I am niclas@apache.org, and an ASF Member with ICLA on file with the secretary.

### DIFF
--- a/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/AssemblyMojo.java
+++ b/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/AssemblyMojo.java
@@ -409,7 +409,7 @@ public class AssemblyMojo extends MojoSupport {
                 case Boot:      bootKars.add(uri); break;
                 case Installed: installedKars.add(uri); break;
                 }
-            } else if ("features".equals(artifact.getClassifier())) {
+            } else if ("features".equals(artifact.getClassifier()) || "karaf".equals(artifact.getClassifier())) {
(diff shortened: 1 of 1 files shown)


jbonofre:
R: @jbonofre

tonit:
Thanks for integrating that fix quickly, @jbonofre. 
Is there a chance to make this more generic? Why depend on classifier string value instead of the xml content or xsd "type" declaration of the content?
If it is a concern about download "unnecessarily", why would someone put in a dependency which is not needed? 
Another issue: with that, the karaf plugin works differently than karaf itself (it uses mvn url handler, i think it is not checking any strings before loading). This made it harder to understand the problem @niclash fixed in the first place.

gnodet:
Right, I think it would make sense to discover the type of artifact (kar / repository / bundle) instead of relying on the artifact type string.  I've just raised https://issues.apache.org/jira/browse/KARAF-5004.
However, I don't understand what you mean by "karaf plugin works differently than karaf itself" ...

gnodet:
Possible patch for KARAF-5004 at https://github.com/gnodet/karaf/commit/801268e654f64803f38661b6bdba2bf884d6b798

hhoffstaette:
@gnodet Very nice, this is more or less how I thought the plugin worked _before_ we came across the bug that @niclash tracked down.

PR #289, opened 2017-03-16
KARAF-5028: set java.io.tmpdir property in karaf-wrapper.conf

johnpoth:
https://issues.apache.org/jira/browse/KARAF-5028


jbonofre:
R: @jbonofre

PR #290, opened 2017-03-22
KARAF-5055: use karaf shutdown timeout when notifying the service wra…

johnpoth:
…pper

https://issues.apache.org/jira/browse/KARAF-5055

Instead of notifying the service wrapper every 5 seconds that we are awaiting shutdown for 5*2 seconds or so, I propose to notify once the service wrapper that we are shutting down using the Karaf shutdown timeout.

I've tested both when the timeout is reached(ungraceful shutdown) and when it's not (graceful shutdown). Both cases yield expected behavior.

Thanks!

### DIFF
--- a/main/src/main/java/org/apache/karaf/main/Main.java
+++ b/main/src/main/java/org/apache/karaf/main/Main.java
@@ -638,11 +638,13 @@ public class Main {
             return true;
         }
         try {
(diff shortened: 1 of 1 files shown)


jbonofre:
R: @jbonofre

PR #291, opened 2017-03-24
[KARAF-4701] forcing deterministic heuristic of capbillities comparison for Subsystem Resolver

vladimirfx:
When some bundle exactly duplicated in two or more installing features Felix Resolver may randomly chose one that not already installed. 
Karaf already use some hack for deterministic resolution in CandidateComparator. But as now it can't produce stable result for absolutly identical bundles.
This patch expand deterministic resolution hack to use Requirement identity in comparison, ie compare from witch feature/subsystem/region bundle is coming.
Steps to reproduce described in bug report. All version from 4.0.6 up to current master is affected. But with Felix Resolver 1.12.0 it is harder to reproduce.


jbonofre:
R: @jbonofre

PR #294, opened 2017-04-13
Remove the VM option "-server"

earcam:
Hi,

If not explicitly defined, the vm option `-server` has been 
automatically set since Java 5.  The bar for it picking `-server` over 
`-client` is pretty low by modern standards; 2 CPU and 2GB RAM.

[http://docs.oracle.com/javase/8/docs/technotes/guides/vm/server-class.html](http://docs.oracle.com/javase/8/docs/technotes/guides/vm/server-class.html)

The Oracle ARMv7 HFP does not support `-server`.  So when running on a 
Raspberry Pi, one must either set JAVA_OPTS or regex replace the 
relevant lines in launch scripts.

IMO it would be better to remove this _redundant_ flag, as I'd guess 
there are few systems that would actually require this to be explicitly 
set.

cheers,
Caspar

### DIFF
--- a/assemblies/features/base/src/main/filtered-resources/resources/bin/inc
+++ b/assemblies/features/base/src/main/filtered-resources/resources/bin/inc
@@ -274,7 +274,7 @@ setupDefaults() {
                 DEFAULT_JAVA_OPTS="${DEFAULT_JAVA_OPTS} -XX:MaxPermSize=${JAVA_MAX_PERM_MEM}"
             fi
         fi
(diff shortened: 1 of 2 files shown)


jbonofre:
R: @jbonofre

earcam:
Thank you =)

jbonofre:
Just for the record, I'm creating a Jira. It will be include in the Release Notes. I will amend your commit with the Jira ID.

earcam:
Sure thing, thanks JB.

PR #301, opened 2017-05-15
Extract Regions from FeaturesServiceImpl

cschneider:


features/core/src/main/java/org/apache/karaf/features/internal/service/FeaturesServiceImpl.java
```diff
@@ -122,13 +101,6 @@
...
     private final Bundle bundle;
     private final BundleContext bundleContext;
-
-    /**
-     * The system bundle context.
-     * For all bundles related operations, we use the system bundle context
-     * to allow this bundle to be stopped and still allow the deployment to
-     * take place.
-     */
     private final BundleContext systemBundleContext;
```
gnodet: Why the removal of this javadoc ?  It looks important to me.

features/core/src/main/java/org/apache/karaf/features/internal/service/FeaturesServiceImpl.java
```diff
@@ -234,14 +181,14 @@ public FeaturesServiceImpl(Bundle bundle,
                                String blacklisted,
                                boolean configCfgStore) {
         this.bundle = bundle;
-        this.bundleContext = bundleContext;
+        this.bundleContext = bundle == null ? null : bundle.getBundleContext();
```
gnodet: I'd rather keep the bundleContext as a parameter.
The reason is that bundle.getBundleContext() can possibly throw an exception.

features/core/src/main/java/org/apache/karaf/features/internal/service/Regions.java
```diff
@@ -0,0 +1,62 @@
...
+    void resolveBundles(Set<Bundle> bundles, Map<Resource, List<Wire>> wiring,
+                        Map<Resource, Bundle> resToBnd);
+
+    void replaceDigraph(Map<String, Map<String, Map<String, Set<String>>>> policies,
+                        Map<String, Set<Long>> bundles)
+        throws BundleException, InvalidSyntaxException;
+
+    void saveState() throws IOException;
+
+    RegionDigraph getGraph() throws BundleException;
```
gnodet: Maybe rename to getDigraphCopy() to better reflect what it actually does ?

features/core/src/main/java/org/apache/karaf/features/internal/service/Regions.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Set;
+
+import org.eclipse.equinox.region.RegionDigraph;
+import org.osgi.framework.Bundle;
+import org.osgi.framework.BundleException;
+import org.osgi.framework.InvalidSyntaxException;
+import org.osgi.resource.Resource;
+import org.osgi.resource.Wire;
+
+public interface Regions {
```
gnodet: `Regions` does not sound a very good name to me...  Too vague.  What about something like `RegionInstallationSupport` ? Another thought ?

features/core/src/main/java/org/apache/karaf/features/internal/service/Deployer.java
```diff
@@ -110,26 +110,12 @@
 public class Deployer {
 
     public interface DeployCallback {
-
         void print(String message, boolean verbose);
-
         void saveState(State state);
         void persistResolveRequest(DeploymentRequest request) throws IOException;
         void installFeature(Feature feature) throws IOException, InvalidSyntaxException;
```
gnodet: I think it would make sense to move the `installFeature` too, as the `DeployCallback` would then only contain state and listeners management.

features/core/src/main/java/org/apache/karaf/features/internal/service/FeaturesServiceImpl.java
```diff
@@ -191,37 +163,12 @@
...
-                               String blacklisted) {
-        this(bundle, bundleContext,systemBundleContext, storage, featureFinder, configurationAdmin,
-                resolver, digraph, overrides, featureResolutionRange, bundleUpdateRange, updateSnaphots,
-                serviceRequirements, globalRepository, downloadThreads, scheduleDelay, scheduleMaxRun, blacklisted,
-                FeaturesService.DEFAULT_CONFIG_CFG_STORE);
-    }
-
-    public FeaturesServiceImpl(Bundle bundle,
-                               BundleContext bundleContext,
                                BundleContext systemBundleContext,
```
gnodet: Can me split out the removal of this constructor in a different commit ?

tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/VerifyMojo.java
```diff
@@ -105,7 +107,8 @@
 import org.osgi.resource.Wire;
 import org.osgi.service.resolver.ResolutionException;
 
-import static java.util.jar.JarFile.MANIFEST_NAME;
+import aQute.bnd.osgi.Macro;
+import aQute.bnd.osgi.Processor;
```
gnodet: Can we avoid reordering imports ?

features/core/src/main/java/org/apache/karaf/features/internal/service/FeaturesServiceImpl.java
cschneider: I moved the comment to Regions as I think the functionality refered by the comment is also moved. The remaining use of systemBundleContext is in getDeploymentState and I had the impression the comment does not match the usage there.. btw. as only one method uses systemBundleContext we can maybe move it out of FeaturesServiceImpl in a later refactoring.

features/core/src/main/java/org/apache/karaf/features/internal/service/FeaturesServiceImpl.java
cschneider: ok

features/core/src/main/java/org/apache/karaf/features/internal/service/Regions.java
cschneider: I was also unsure. Not sure if RegionInstallationSupport is better but if you prefer it that is fine for me.

features/core/src/main/java/org/apache/karaf/features/internal/service/Regions.java
cschneider: ok

tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/VerifyMojo.java
cschneider: I will try to avoid it. Sometimes I tend to press Ctrl-O :-)

features/core/src/main/java/org/apache/karaf/features/internal/service/Deployer.java
cschneider: I think installFeature is very similar to uninstallFeature. So we would need to move both. If you look what they depend on there would not be much left in FeaturesServiceImpl. 

I have some ideas how to improve the situation but they would require a bigger redesign.

features/core/src/main/java/org/apache/karaf/features/internal/service/FeaturesServiceImpl.java
cschneider: Why would you like it to be separate?

gnodet: Yeah, smaller commits are easier to review imho

features/core/src/main/java/org/apache/karaf/features/internal/service/FeaturesServiceImpl.java
gnodet: Ok

PR #306, opened 2017-05-31
KARAF-5168: replace old-style loops

skitt:
This patch replaces a number of old-style explicit loops over arrays
or iterators with foreach loops or streams.


jbonofre:
R: @jbonofre

PR #310, opened 2017-06-01
[KARAF-5173] Provide an impl of CallbackHandler

skitt:
A number of tests use an anonymous class implementing CallbackHandler
with the same pattern, username and password. This patch introduces
NamePasswordCallbackHandler and uses it where appropriate.


jbonofre:
R: @jbonofre

jbonofre:
retest this please

skitt:
Karaf passes all its tests on my system with this patch applied (and likewise on Jenkins). Is there something specific that gives you cause for concern?

jbonofre:
No, I'm just testing the connectivity between Jenkins and GitHub. We have an issue that the infra is investigating. Again, nothing wrong with the pull request itself ;)

jbonofre:
By the way "retest this please" is a keyword comment that triggers a Jenkins build on the pull request.

skitt:
Ah right OK, I wasn’t aware of the keyword ;-).

PR #312, opened 2017-06-02
Update provisioning.adoc

davidmichaelkarr:
Fix two different variations of typos that should be "comma-separated".


jbonofre:
R: @jbonofre

PR #313, opened 2017-06-05
KARAF-5179 Fix usage of karaf.restart.jvm in Main

clockard:
This PR fixes an issue with restarting the jvm using the karaf.restart.jvm property

https://issues.apache.org/jira/browse/KARAF-5179


jbonofre:
R: @jbonofre

PR #315, opened 2017-06-07
Openssh keygen fix

llech:
When no SSH key was present, a key was generated. Unfortunately, the write method was incompatibile with read, and such key could not be read on next start, which generated issues such as:
https://stackoverflow.com/questions/43070457/stop-karaf-from-generating-new-ssh-keys-on-startup

I've fixed the key persisting and provided JUnit test that checks if that what was written can be read. Previous 'test' has only tested if an exception was thrown. Because an exception was consumed, the 'test' was passed although the read has failed.


gnodet:
I'll have a look.

jbonofre:
R: @gnodet 
CC: @jbonofre

PR #318, opened 2017-06-12
A fix to KARAF-5104 karaf:run should support a features set

steinarb:
This pull request is a fix to [KARAF-5104](https://issues.apache.org/jira/browse/KARAF-5104)

This change adds:

1. A parameter to set a list of URLs to feature repositories to add to the karaf started by karaf:run
2. A parameter containing a comma separated list of features to install in the karaf started by karaf:run
3. If karaf:run is set up to deploy the current project artifact and if the current project has packaging pom and an attached file with classifier "features" and type "xml", that attachment will be installed as a feature repository in the karaf started by the karaf:run goal

Even if you decide not to take this change, you might consider the first commit of this pull request, because that is just as-is unit tests of the RunMojo.deploy(BundleContext) method.

### DIFF
--- a/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/RunMojo.java
+++ b/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/RunMojo.java
@@ -75,6 +75,13 @@ public class RunMojo extends MojoSupport {
     private boolean deployProjectArtifact = true;
 
     /**
+     * A list of URLs referencing feature repositories that will be added
+     * to the karaf instance started by this goal.
+     */
+    @Parameter
+    private String[] featureRepositories = null;
+
+    /**
      * Comma-separated list of features to install.
      */
     @Parameter(defaultValue = "")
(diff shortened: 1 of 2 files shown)


jbonofre:
R: @jbonofre

tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/RunMojo.java
```diff
@@ -145,12 +185,30 @@ private void deploy(BundleContext bundleContext) throws MojoExecutionException {
...
 
+    void addFeatures(Object featureService) throws MojoExecutionException {
+    	if (featuresToInstall != null) {
+            try {
+            	Class<? extends Object> serviceClass = featureService.getClass();
+                Method installFeatureMethod = serviceClass.getMethod("installFeature", String.class);
+                String[] features = featuresToInstall.split(" *, *");
+                for (String feature : features) {
+                    installFeatureMethod.invoke(featureService, feature);
+                    Thread.sleep(1000L);
```
jbonofre: Why this sleep of 1s ?

jbonofre:
I will update the user guide accordingly.

tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/RunMojo.java
steinarb: I used my own application to test it and my own application triggers [KARAF-5174](https://issues.apache.org/jira/browse/KARAF-5174). In addition to the uninstall problem there is a startup problem seen in the karaf.log if the features are loaded together (the derby driver fails to be able to create the memory database), but if the liquibase-core feature is installed _before_ the database feature there are no such issues in the karaf.log.

I saw something similar in the karaf.log when running karaf:run on my app, so I tried the 1s sleep to simulate what I do manually at the karaf console.  But it didn't make any difference.  So feel free to drop the sleep if it doesn't make sense.

jbonofre: Fair enough, let's keep like this as a first step. We will improve later. Thanks.

PR #320, opened 2017-06-15
[KARAF-5208] Improve feature:install error message

jludvice:
R: @jbonofre

### DIFF
--- a/features/core/src/main/java/org/apache/karaf/features/internal/util/MultiException.java
+++ b/features/core/src/main/java/org/apache/karaf/features/internal/util/MultiException.java
@@ -61,4 +61,15 @@ public class MultiException extends Exception {
             throw new MultiException(sb.toString(), exceptions);
         }
     }
+
+    @Override
+    public String getMessage() {
+        StringBuilder sb = new StringBuilder(super.getMessage());
+        sb.append(":");
+        for (Throwable e : getSuppressed()) {
+            sb.append("\n\t");
+            sb.append(e.getMessage());
+        }
+        return sb.toString();
+    }
 }


jbonofre:
Thanks ! I'm reviewing.

gnodet:
Shouldn't we override ```getMessage()``` instead ?
Something like:
```
public String getMessage() {
            StringBuilder sb = new StringBuilder(super.getMessage());
            sb.append(":");
            for (Exception e : exceptions) {
                sb.append("\n\t");
                sb.append(e.getMessage());
            }
}
```

jludvice:
On which class @gnodet ?

gnodet:
On ```MultiException```.  It should then print the root causes, even if the full stack is not printed.

jludvice:
This also works for me.
```
karaf@root()> feature:install spring-jdbc                                                                                                                               15:13:00
Error executing command: Error:
        Error downloading mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.spring-jdbc/4.3.7.RELEASE_1
```

compared to previous
```
karaf@root()> feature:install spring-jdbc
Error executing command: Error downloading mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.spring-tx/4.3.7.RELEASE_1
```

@gnodet's solution looks less hackish.
Are we going to prefer overriding `getMessage`?

jbonofre:
Personally, I don't mind: either way is fine to me.

jludvice:
Just from curiosity, why is this class implemented three times in karaf?
* https://github.com/apache/karaf/blob/master/features/core/src/main/java/org/apache/karaf/features/internal/util/MultiException.java
* https://github.com/apache/karaf/blob/master/shell/console/src/main/java/org/apache/karaf/shell/console/MultiException.java
* https://github.com/apache/karaf/blob/master/shell/core/src/main/java/org/apache/karaf/shell/support/MultiException.java

Wouldn't it make sense to have just one where we include this fix?

gnodet:
Overriding the ```getMessage``` has the benefit of displaying all the inner exceptions in case there are more than one problem.  The previous solution only displays the first one.

jbonofre:
@gnodet that makes sense. @jludvice can you please update accordingly ?

jbonofre:
retest this please

jludvice:
retest this please

jbonofre:
retest this please

jbonofre:
I updated Jenkins config to be able to trigger pull request build.

PR #322, opened 2017-06-21
Master maven commands

grgrzybek:


grgrzybek:
@gnodet @jbonofre @cschneider please review if you like

None of your `~/.m2/settings.xml` or `~/.m2/settings-security.xml` will break

jbonofre:
R: @jbonofre

grgrzybek:
ah - this new bundle isn't installed by default and there's no feature for it (yet). You have to install it manually: `install -s mvn:org.apache.karaf.maven/org.apache.karaf.maven.core/4.2.0-SNAPSHOT`

jbonofre:
I suggest to add a "maven" feature for convenience.

PR #327, opened 2017-07-04
Fixed authentication when userInfo provided with resource URL

dleangen:
In a features file, when the "userInfo" (username:password) is provided with the URL, the authentication is not getting processed.

This patch takes the userInfo and creates an HTTP header for authentication.

Note: I am not sure if the userInfo should be removed from the URL or not before the request is made, so I just left it intact.

### DIFF
--- a/features/core/src/main/java/org/apache/karaf/features/internal/repository/UrlLoader.java
+++ b/features/core/src/main/java/org/apache/karaf/features/internal/repository/UrlLoader.java
@@ -20,7 +20,9 @@ import java.io.BufferedInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.HttpURLConnection;
+import java.net.URL;
 import java.net.URLConnection;
+import java.nio.charset.StandardCharsets;
 import java.util.zip.GZIPInputStream;
 
 import static java.net.HttpURLConnection.HTTP_NOT_MODIFIED;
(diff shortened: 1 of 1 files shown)


jbonofre:
R: @jbonofre

jbonofre:
I will create the corresponding Jira (for the tracking in Release Notes) and amend your commit message with the Jira ID. Thanks !

jbonofre:
@dleangen I saw your directly commit your changes. Please, next time, consider my comments, thanks ;)

dleangen:
Hi @jbonofre, which comments are you referring to? Did I not do the PR correctly?

Sorry, I'm not sure what you mean. It was not my intent to do anything improperly. Could you please explain a little more what you mean?

cschneider:
Hi David,
ideally you should create a jira issue describing the problem before doing a PR. Then in the commit(s) you should use the jira id like [KARAF-123] Your message...
You can then also name the branch like the jira issue id. (Not mandatory as the branch is lost after the merge anyway).
So the jira issue is part of the next release notes and the commit is linked to the jira issue.
Christian

We can also do that by amending the commit message of course but it is easier if you already provide this information.

dleangen:
Ok, got it! Thanks for explaining.

cschneider:
I will add this information to the community docs so we can refer to it and you have a nice reference to look up. See https://karaf.apache.org/community.html

dleangen:
:-)

PR #338, opened 2017-08-29
shell.core instead of shell.console

vorburger:
The @org.apache.karaf.shell.commands.Command from org.apache.karaf.shell.console is deprecated, so wouldn't it be better to document using the org.apache.karaf.shell.core which has org.apache.karaf.shell.api.action.Command ?


vorburger:
@cschneider @skitt ^^^

jbonofre:
R: @jbonofre

PR #345, opened 2017-09-05
Primitives should not be boxed just for Strign conversion

apupier:
Signed-off-by: Aurélien Pupier <apupier@redhat.com>

### DIFF
--- a/webconsole/gogo/src/main/java/org/apache/karaf/webconsole/gogo/Terminal.java
+++ b/webconsole/gogo/src/main/java/org/apache/karaf/webconsole/gogo/Terminal.java
@@ -1025,7 +1025,7 @@ public class Terminal {
         String prefix = "";
         if (p.length() > 0) {
             if (p.charAt(0) >= '<' && p.charAt(0) <= '?') {
-                prefix = "" + p.charAt(0);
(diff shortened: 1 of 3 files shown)


jbonofre:
R: @jbonofre

webconsole/gogo/src/main/java/org/apache/karaf/webconsole/gogo/Terminal.java
```diff
@@ -1710,14 +1710,14 @@ public synchronized String pipe(String d) {
                 vt100_keyfilter_escape = true;
             } else if (c == 127) {
                 if (vt100_mode_backspace) {
-                    o += (char) 8;
+                    o += Integer.toString(8);
```
jbonofre: It's a `String` whereas we cast to `char`.

apupier: yes but the char is then autocasted to String. I think that we can get rid of this step, isn't it?

jbonofre: Agree, it's just a personal note for the merge ;)

PR #346, opened 2017-09-05
EntrySet should be iterated when both the key and value are needed

apupier:
Signed-off-by: Aurélien Pupier <apupier@redhat.com>

### DIFF
--- a/bundle/core/src/main/java/org/apache/karaf/bundle/command/ShowBundleTree.java
+++ b/bundle/core/src/main/java/org/apache/karaf/bundle/command/ShowBundleTree.java
@@ -139,15 +139,16 @@ public class ShowBundleTree extends BundleCommand {
                 }
             }
         }
-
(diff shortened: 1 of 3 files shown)


jbonofre:
R: @jbonofre

apupier:
it brings performance value, it avoids to access and search two times in the map. it is making a visible effect for large Maps but i don't know which ones can be big or not.

Please note that there are a lot more of this in the codebase.

jbonofre:
FYI, the compiler (since Java 1.7 update 40) can already optimize the iteration on the map. Further more, those maps are pretty small ;) Thanks anyway !

apupier:
@jbonofre do you have a reference link for that? i would like to report it to Sonar in order that they don't report it as an issue if source is requiring Java 7

jbonofre:
Let me find it. Anyway, I prefer your approach, it's clearer ;)

PR #367, opened 2017-09-14
Use Map.computeIfAbsent() in map population

rovarga:
Instead of performing two lookups, perform a computeIfAbsent(),
which perform a fused lookup-or-put.


jbonofre:
R: @jbonofre

PR #368, opened 2017-09-14
Speed up string splitting operations

rovarga:
Instead of performing multiple String.indexOf(String) operations,
use String.indexOf(char) and cache its result.

### DIFF
--- a/features/core/src/main/java/org/apache/karaf/features/internal/model/Feature.java
+++ b/features/core/src/main/java/org/apache/karaf/features/internal/model/Feature.java
@@ -128,9 +128,9 @@ public class Feature extends Content implements org.apache.karaf.features.Featur
 
     public static org.apache.karaf.features.Feature valueOf(String str) {
         if (str.contains(VERSION_SEPARATOR)) {
-            String strName = str.substring(0, str.indexOf(VERSION_SEPARATOR));
-            String strVersion = str.substring(str.indexOf(VERSION_SEPARATOR)
-                    + VERSION_SEPARATOR.length(), str.length());
+            int idx = str.indexOf(VERSION_SEPARATOR);
+            String strName = str.substring(0, idx);
+            String strVersion = str.substring(idx + VERSION_SEPARATOR.length(), str.length());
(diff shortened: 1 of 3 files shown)


gnodet:
What about switching the `VERSION_SEPARATOR` to a `char` ?

features/core/src/main/java/org/apache/karaf/features/internal/model/Feature.java
```diff
@@ -128,9 +128,9 @@ public Feature(String name, String version) {
 
     public static org.apache.karaf.features.Feature valueOf(String str) {
         if (str.contains(VERSION_SEPARATOR)) {
-            String strName = str.substring(0, str.indexOf(VERSION_SEPARATOR));
-            String strVersion = str.substring(str.indexOf(VERSION_SEPARATOR)
-                    + VERSION_SEPARATOR.length(), str.length());
+            int idx = str.indexOf(VERSION_SEPARATOR);
```
gnodet: You should switch `VERSION_SEPARATOR` to a `char` and use:
```
int idx = str.indexOf(VERSION_SEPARATOR);
if (idx >= 0) {
```
instead.

features/core/src/main/java/org/apache/karaf/features/internal/service/FeaturesServiceImpl.java
```diff
@@ -892,13 +892,18 @@ private State copyState() {
...
-    private String normalize(String feature) {
-        if (!feature.contains(VERSION_SEPARATOR)) {
-            feature += "/0.0.0";
+    private static String normalize(String feature) {
+        final int idx = feature.indexOf(VERSION_SEPARATOR);
+        final String name;
+        final String version;
+        if (idx == -1) {
+            name = feature;
+            version = "0.0.0";
```
gnodet: You could return here to avoid parsing / cleaning the "0.0.0" version.

gnodet:
The changes have been merged, thx for the patch.
Could you please close this PR now ?

PR #374, opened 2017-09-18
[KARAF-5315] Synchronize lock methods

jbonofre:


jbonofre:
retest this please

main/src/main/java/org/apache/karaf/main/lock/SimpleFileLock.java
```diff
@@ -80,29 +80,12 @@ public void release() throws Exception {
...
-            if (path != null) {
-                rc = validateDirectoryExists(path, "Invalid " + PROP_KARAF_BASE + " property");
-            }
-        }
-
-        if (rc == null) {
-            rc = lock;
-        }
-        return rc;
+        return validateDirectoryExists(lock.getPath(), "Invalid" + PROPERTY_LOCK_DIR + " system property");
```
gnodet: I think the code behaves differently if {{PROPERTY_LOCK_DIR}} is not set.
Previously, it would default to {{KARAF_BASE}}, now it returns null and will certainly cause a NPE in the constructor.

jbonofre: Good point, let me change this part and keep the sync on the methods.

jbonofre: Thanks !

jbonofre:
Squashed the commits. Waiting the end of Jenkins before merging.

jbonofre:
retest this please

PR #376, opened 2017-09-26
Turn Feature.VERSION_SEPARATOR into a char

rovarga:
String.indexOf(char) is faster than String.indexOf(String), so use
char for VERSION_SEPARATOR.


jbonofre:
R: @jbonofre

PR #377, opened 2017-09-26
Clean up object duplication

rovarga:
This series of patches are attempts to improve resolver memory usage, which we have
observed to go up to ~2GB while performing a mass install of all OpenDaylight features.

A large-ish part of this are duplicated Strings and Maps.


jbonofre:
R: @jbonofre

PR #379, opened 2017-10-02
KARAF-5397: Remove config for org.apache.karaf.shell.cfg from ssh fea…

graben:
…ture

### DIFF
--- a/assemblies/features/standard/src/main/feature/feature.xml
+++ b/assemblies/features/standard/src/main/feature/feature.xml
@@ -663,12 +663,6 @@
     </feature>
 
     <feature name="ssh" description="Provide a SSHd server on Karaf" version="${project.version}">
-        <config name="org.apache.karaf.shell">
-            sshPort=8101
-            sshHost=0.0.0.0
-            sshRealm=karaf
-            hostKey=${karaf.etc}/host.key
-        </config>
         <feature>shell</feature>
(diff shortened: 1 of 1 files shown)


jbonofre:
R: @jbonofre

graben:
Retest this please

graben:
Something wrong with jenkins? Tried to build twice but without success.

jbonofre:
Yes, there's a problem on Jenkins (not related to your PR):

```
java.io.IOException: failed to create temp directory at default location, most probably at: /tmp
```

I gonna deal to fix that.

assemblies/features/standard/src/main/feature/feature.xml
```diff
@@ -663,12 +663,6 @@
     </feature>
 
     <feature name="ssh" description="Provide a SSHd server on Karaf" version="${project.version}">
-        <config name="org.apache.karaf.shell">
```
jbonofre: This config is used by the default distribution, so I'm not sure it makes sense to remove this. I would update it instead.

graben: Well, shell is also used by the default distribution and it also sets this config but actually gets overridden by ssh.

jbonofre: Got it, let me test the PR.

PR #380, opened 2017-10-03
KARAF-5403 Fix for loosing order of features given in tooling configuration

splatch:


gnodet:
Feel free to go ahead and commit it directly if you want.

PR #381, opened 2017-10-06
[KARAF-5411] Allow for users.properties file not existing, or empty

tomq42:


jbonofre:
Please, close this PR. The PR should be opened only on master and then we cherry-pick on other branches.

tomq42:
Fair enough.
The code was significantly different in the Config class, and I actually couldn't get master to run, and 4.1 was what I was after.

PR #382, opened 2017-10-06
[KARAF-5411] Allow for users.properties file not existing, or empty

tomq42:


jbonofre:
R: @jbonofre

tomq42:
What's the process here? What can I do to expedite this getting included?
I ultimately want a version of 4.1.3 with this change in it (well, almost this change, can't be exactly the same). Currently we're building something ourselves, but I'd rather use the official snapshot builds.
Thanks.

jbonofre:
I'm in the process of review. Once the review is OK, according to the Jira, I will merge on master and cherry-pick on 4.1.x branch. Just give me time for the review ;)

tomq42:
OK, no problem. 
I only asked because we were having problems our end with our build process. I have naively built a version of 4.1.3-SNAPSHOT with this change in, but of course as soon as there's an "official" 4.1.3-SNAPSHOT that's later than ours, that gets picked up instead, so it's random whether we have this fix in or not. Rather than working out how to set up our artifactory or maven to always take our snapshot rather than yours, I thought the simplest thing would be if this patch was in the official snapshot builds.

client/src/main/java/org/apache/karaf/client/Main.java
```diff
@@ -142,14 +142,28 @@ public String getUpdatedPassword(ClientSession session, String prompt, String la
                     }
                 });
             }
+            
+            if (config.getUser()==null || config.getUser().isEmpty()) {
+            	while (true) {
+            		String user = console.readLine("Enter user : ");
```
jbonofre: In order to be consistent with the password prompt, I will use `"Enter user:  "`.

tomq42:
Cool, Thnx.

PR #383, opened 2017-10-09
KARAF-5413 - Added missing version in features.xml descriptors

paoloantinori:


assemblies/features/standard/src/main/feature/feature.xml
```diff
@@ -560,7 +560,7 @@
         <capability>http-service;provider:=felix-httplite</capability>
     </feature>
 
-    <feature name="pax-http-service" description="Pax-Web OSGi HTTP Service">
+    <feature name="pax-http-service" description="Pax-Web OSGi HTTP Service" version="${project.version}">
```
jbonofre: Not sure it's a good idea to use `${project.version}`.  I would have used `${pax.web.version}`.

assemblies/features/standard/src/main/feature/feature.xml
```diff
@@ -972,7 +972,7 @@
         <bundle>mvn:org.apache.aries.blueprint/org.apache.aries.blueprint.webosgi/${aries.blueprint.web.version}</bundle>
     </feature>
 
-    <feature name="wrap" description="Wrap URL handler">
+    <feature name="wrap" description="Wrap URL handler" version="${project.version}">
```
jbonofre: Same here, I would have used the Pax URL version instead of the Karaf version.

PR #384, opened 2017-10-11
ENTESB-64: Shutting down Fuse ESB does not delete lock file

RaiSaurabh:
Updated SimpleFileLock class to delete the lock file after release method is called for releasing lock.


gnodet:
I'm not sure this patch is good.  I've done a bit of testing and this breaks the locking mechanism.
If you don't need locking, it would be easier to disable it using `karaf.lock = false` in `etc/config.properties`.

jbonofre:
Yeah, agree, it doesn't look to me neither as an user might want to cleanly stop an instance without removing the lock file to let a slave instance to startup.

PR #398, opened 2017-11-21
KARAF-5496 - NPEs in SyncopeLoginModule if "version" is not specified

coheigea:


jbonofre:
By the way, the build fails on `EventLoggerTest`. I will fix that. No need to do the rebase (I will test on another PR).

PR #416, opened 2017-12-07
[KARAF-5528] Stop refreshed bundles together with updated and deleted…

grgrzybek:
… ones

### DIFF
--- a/features/core/src/main/java/org/apache/karaf/features/internal/service/Deployer.java
+++ b/features/core/src/main/java/org/apache/karaf/features/internal/service/Deployer.java
@@ -559,7 +559,7 @@ public class Deployer {
         //
         // Execute deployment
         //
-        // #1: stop bundles that needs to be updated or uninstalled in order
+        // #1: stop bundles that needs to be updated or uninstalled or refreshed in order
         // #2: uninstall needed bundles
         // #3: update regions
         // #4: update bundles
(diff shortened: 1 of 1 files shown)


grgrzybek:
@gnodet I checked with fabric8 itests. in this case I had problem with too early stopping of `dstate.serviceBundle`, thus this slight modification. Please verify

grgrzybek:
@gnodet I checked with fabric8 itests. in this case I had problem with too early stopping of `dstate.serviceBundle`, thus this slight modification. Please verify

grgrzybek:
retest this please

features/core/src/main/java/org/apache/karaf/features/internal/service/Deployer.java
```diff
@@ -633,6 +633,9 @@ public void deploy(DeploymentState dstate, DeploymentRequest request) throws Exc
             toStop.addAll(regionDeployment.toUpdate.keySet());
             toStop.addAll(regionDeployment.toDelete);
         }
+        Set<Bundle> toRefreshToStopEarly = new HashSet<>(toRefresh.keySet());
+        toRefreshToStopEarly.remove(dstate.serviceBundle);
+        toStop.addAll(toRefreshToStopEarly);
```
gnodet: Is there a need for this intermediary collection ?
What about 
```
toStop.addAll(toRefresh.keySet());
toStop.remove(serviceBundle);
```

gnodet:
I thought the features service could support being stopped.
Do you recall what the problem was ?

features/core/src/main/java/org/apache/karaf/features/internal/service/Deployer.java
grgrzybek: intermediary was for case, when initial `toStop` contained service bundle. I didn't want to change that, as it worked.

grgrzybek:
yes - stopping service bundle was supported. But in fabric8 case (with karaf 4 resolver), when toRefresh contained serviceBundle, I didn't want it to stop too early, as the process used it in between - it's probably not a problem in current Karaf4, but it *is* in fabric8v1

features/core/src/main/java/org/apache/karaf/features/internal/service/Deployer.java
gnodet: That's the part I don't get.
If it was supported, I don't see why you had a problem.  if it was not, we should not keep it.

PR #439, opened 2018-01-31
KARAF-5585 Do not provide maven settings for pax url if they are gone.

splatch:
This commit rollbacks/moves changes made in KARAF-4288 to verify mojo.


jbonofre:
For the record, it's better to open a branch based on `master`

PR #446, opened 2018-02-09
[KARAF-5604] Speed up descriptor generation

rovarga:
Profiling has turned out two more inefficiencies:
- calling log.debug() with a concatenated string -- we end up wasting time constructing the string, which is never logged
- aether's DefaultArtifact(String) is extremely slow due to Pattern.compile() it contains


jbonofre:
retest this please

PR #462, opened 2018-02-20
Fixes karaf:run goal for single bundle execution

jozzan:
See issue: https://issues.apache.org/jira/browse/KARAF-5630

### DIFF
--- a/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/RunMojo.java
+++ b/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/RunMojo.java
@@ -99,6 +99,10 @@ public class RunMojo extends MojoSupport {
     @Parameter(defaultValue = "false")
     private String startSsh = "false";
 
+
+    @Parameter(defaultValue="${project.build.directory}/${project.build.finalName}.jar")
+    private String bundleArchivePath;
+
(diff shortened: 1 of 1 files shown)


gnodet:
retest this please

tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/RunMojo.java
```diff
@@ -141,6 +145,7 @@ public void execute() throws MojoExecutionException, MojoFailureException {
 
             Object featureService = findFeatureService(bundleContext);
             addFeatureRepositories(featureService);
+            fixProjectArtifact();
```
jbonofre: I would avoid to introduce such method name here. Why not simply adding a comment with the code inline ?

tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/RunMojo.java
```diff
@@ -153,6 +158,14 @@ public void execute() throws MojoExecutionException, MojoFailureException {
         }
     }
 
+    void fixProjectArtifact() {
+        Artifact artifact = project.getArtifact();
+
+        if(artifact.getFile() == null || !artifact.getFile().exists()) {
+            artifact.setFile(new File(bundleArchivePath));
```
jbonofre: Sorry, but I didn't understand the purpose here. `karaf:run` is supposed to be in an execution cycle, so it should not attach any artifact.

jbonofre:
See my comment in Jira, it already works as expected.

PR #466, opened 2018-02-23
DeployMojo fix ssh deploy

jozzan:
issue: [5642](https://issues.apache.org/jira/browse/KARAF-5642)

### DIFF
--- a/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/client/DeployMojo.java
+++ b/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/client/DeployMojo.java
@@ -199,7 +199,7 @@ public class DeployMojo extends MojoSupport {
                 print.println("bundle:install -s " + location);
             }
 
-            final ClientChannel channel = session.createChannel("exec", print.toString().concat(NEW_LINE));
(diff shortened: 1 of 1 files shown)


tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/client/DeployMojo.java
```diff
@@ -199,7 +199,7 @@ public String getUpdatedPassword(ClientSession session, String prompt, String la
                 print.println("bundle:install -s " + location);
             }
 
-            final ClientChannel channel = session.createChannel("exec", print.toString().concat(NEW_LINE));
+            final ClientChannel channel = session.createChannel("exec", writer.toString().concat(NEW_LINE));
```
jbonofre: Why using the writer whereas we have the printwriter ?

```
            StringWriter writer = new StringWriter();
            PrintWriter print = new PrintWriter(writer, true);
```

jozzan: Well the PrintWriter doesn't override the toString method.

jbonofre: Ok got it. So the consequence is a invalid string. Thx.

PR #467, opened 2018-02-28
[KARAF-5647] start, stop, shell, status and client fail on Solaris Sp…

oscerd:
…arc 11


oscerd:
@jbonofre build is good.

PR #468, opened 2018-03-02
[KARAF-5649] Add a feature for the manual

fpapon:
Adding a documentation feature in order to published the generated manual in html with the http service.

### DIFF
--- a/assemblies/features/standard/src/main/feature/feature.xml
+++ b/assemblies/features/standard/src/main/feature/feature.xml
@@ -1069,6 +1069,11 @@
         </config>
     </feature>
 
+    <feature name="documentation" description="Documentation of Karaf project in HTML" version="${project.version}">
+        <feature>war</feature>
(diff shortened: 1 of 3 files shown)


jbonofre:
The `documentation` feature doesn't validate, it's because the `manual` module is built only with the `release` profile.

jbonofre:
retest this please

jbonofre:
retest this please

jbonofre:
As the `manual` profile is defined after the `assemby`, the error is "expected".

I propose the following:

1. Remove the `manual` profile in the main `pom.xml`
2. Include `manual` module in the main `pom.xml` (reactor) before the `assembly`

It means the manual will be systematically built, which is not bad IMHO. Especially, the PDF generation is isolated in a dedicated profile.

fpapon:
I removed the manual profile from pom parent and include the manual module in the reactor.

Ready for the review ;)

PR #469, opened 2018-03-05
Deploy with no artifactLocations fix

jozzan:
deploy fails with a NullPointer when there are no artifactLocations defined even though useProjectArtifact is set to true.


jbonofre:
Sorry, already fix that.

PR #480, opened 2018-03-27
[KARAF-5678] Do not overwrite existing config files from boot/startup…

grgrzybek:
… features in some cases

### DIFF
--- a/profile/src/main/java/org/apache/karaf/profile/assembly/AssemblyDeployCallback.java
+++ b/profile/src/main/java/org/apache/karaf/profile/assembly/AssemblyDeployCallback.java
@@ -154,45 +154,56 @@ public class AssemblyDeployCallback extends StaticInstallSupport implements Depl
         // Install
         Downloader downloader = manager.createDownloader();
         for (Config config : ((Feature) feature).getConfig()) {
+            Path configFile = etcDirectory.resolve(config.getName() + ".cfg");
+            if (Files.exists(configFile) && !config.isAppend()) {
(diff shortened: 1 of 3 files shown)


profile/src/main/java/org/apache/karaf/profile/assembly/AssemblyDeployCallback.java
```diff
@@ -154,45 +154,56 @@ public void installConfigs(org.apache.karaf.features.Feature feature) throws IOE
         // Install
         Downloader downloader = manager.createDownloader();
         for (Config config : ((Feature) feature).getConfig()) {
+            Path configFile = etcDirectory.resolve(config.getName() + ".cfg");
+            if (Files.exists(configFile) && !config.isAppend()) {
```
jbonofre: Maybe, we can also introduce `config.isOverride()` to completely replace a existing configuration with the one contained in the feature (as we have for `<configFile/>`. Tought ?

grgrzybek: it'd require XSD version change. adding explicit `override` would be probably confusing - at runtime, when feature resolver installs (or even refreshes) a feature, it would have **not** to overwrite it anyway

jbonofre: OK fair enough. Np

jbonofre:
Maybe wait Jenkins green light before merging.

grgrzybek:
sure - waiting...

PR #486, opened 2018-04-09
[KARAF-3235] Expose karaf itests as a bundle in order to simplify the way of writing itests for end-users

jbonofre:
… way of writing itests for end-users


jbonofre:
retest this please

jbonofre:
retest this please

jbonofre:
`itests` are failing with this change. I'm checking why.

jbonofre:
retest this please

jbonofre:
retest this please

jbonofre:
The `ResourceBuilder` is now part of Felix Utils. Unfortunately the fix I did when `ResourceBuilder` was part of Karaf has not been ported to Felix Utils 1.11.0. I'm fixing on Felix Utils.

jbonofre:
OK, I understand the problem: the `maven-bundle-plugin` is executed *after* the `maven-surefire-plugin` (for the test). So, the MANIFEST doesn't contain the OSGi headers, that's why tests are failing.
The fix is that the `maven-bundle-plugin` has to be executed before `maven-surefire-plugin`. A possible workaround is to split itests in two modules: the bundle (containing `KarafTestSupport`) and the actual itests (as a jar module). I'm doing the change.

PR #489, opened 2018-04-18
Enable karaf-assembly packaging to work with custom deployers

hbfernandes:
Allow the use of the karaf-assembly packaging when using the maven-karaf-plugin to properly add bundles with custom deployers to the repo.

A bundle defined as follows for example will be correctly added to the system folder:

```<bundle>my-deployer:mvn:org.webjars.bower/angular/1.5.8</bundle>```

Please let me know if this causes any issues that I might not be aware of.

### DIFF
--- a/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/features/InstallKarsMojo.java
+++ b/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/features/InstallKarsMojo.java
@@ -562,6 +562,11 @@ public class InstallKarsMojo extends MojoSupport {
             if (bundleLocation.startsWith("war:")) {
                 bundleLocation = bundleLocation.substring("war:".length());
             }
+            // checked all known prefixes, clean a possible custom one wrapping mvn:
+            int mvnIndex = bundleLocation.indexOf("mvn:");
+            if (mvnIndex > 0) {
+                bundleLocation = bundleLocation.substring(mvnIndex);
+            }
             File bundleFile;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/features/InstallKarsMojo.java [hunk 1]
```diff
@@ -562,6 +562,11 @@ private void installBundle(Bundle bundle) throws Exception {
             if (bundleLocation.startsWith("war:")) {
                 bundleLocation = bundleLocation.substring("war:".length());
             }
+            // checked all known prefixes, clean a possible custom one wrapping mvn:
+            int mvnIndex = bundleLocation.indexOf("mvn:");
+            if (mvnIndex > 0) {
+                bundleLocation = bundleLocation.substring(mvnIndex);
+            }
             File bundleFile;
             if (bundleLocation.startsWith("mvn:")) {
```
jbonofre: With your change, it means that this is never reach. Can you add a rationale about this change ? I'm not sure I understand exactly the purpose.

hbfernandes: It will reach it, because the `mvn:` part of the location will remain. All it did was to strip away the custom wrapper.

So `my-deployer:mvn:org.webjars.bower/angular/1.5.8` will become `mvn:org.webjars.bower/angular/1.5.8` making it eligible to enter the other if conditions.

PR #491, opened 2018-04-27
Install bundles with custom deployers and edit featuresBoot

hbfernandes:
- Enable bundles with custom deployers to be installed. 
- Enable assembly property edits to change featuresBoot by moving the application of the properties to the end.

Will any of these changes cause issues?

### DIFF
--- a/profile/src/main/java/org/apache/karaf/profile/assembly/Builder.java
+++ b/profile/src/main/java/org/apache/karaf/profile/assembly/Builder.java
@@ -559,15 +559,6 @@ public class Builder {
             Files.write(configFile, config.getValue());
         }
 
-        // 'improve' configuration files.
-        if (propertyEdits != null) {
-            KarafPropertiesEditor editor = new KarafPropertiesEditor();
-            editor.setInputEtc(etcDirectory.toFile())
-                    .setOutputEtc(etcDirectory.toFile())
-                    .setEdits(propertyEdits);
-            editor.run();
-        }
-
         //
         // Handle overrides
         //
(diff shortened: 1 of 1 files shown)


profile/src/main/java/org/apache/karaf/profile/assembly/Builder.java
```diff
@@ -617,6 +608,15 @@ public String getValue(String key) {
         // Installed stage
         //
         installStage(installedProfile, allBootFeatures);
+
```
jbonofre: For the record, this is already fixed on `master` and `karaf-4.1.x` branches.

profile/src/main/java/org/apache/karaf/profile/assembly/Builder.java
```diff
@@ -1023,6 +1023,11 @@ private Profile startupStage(Profile startupProfile) throws Exception {
         for (Map.Entry<Integer, Set<String>> entry : invertedStartupBundles.entrySet()) {
             String startLevel = Integer.toString(entry.getKey());
             for (String location : new TreeSet<>(entry.getValue())) {
+                // remove any custom deployer prefix
+                int mvnIndex = location.indexOf("mvn:");
```
jbonofre: Same question as in the other PR: I don't understand why you remove the `mvn` prefix from the bundle location.

hbfernandes: Like i said on the other PR, `mvn:` is not being removed.

profile/src/main/java/org/apache/karaf/profile/assembly/Builder.java
hbfernandes: Ok, thats good. Still needed this part in to make my scenario work though.

PR #504, opened 2018-05-14
[KARAF-5753]

StephaneRicci:


assemblies/features/base/src/main/filtered-resources/resources/bin/karaf
```diff
@@ -93,7 +93,12 @@ checkRootInstance() {
...
           then
               MAIN=org.apache.karaf.main.Main
-              PID_COMMAND=$(ps -p "${ROOT_INSTANCE_PID}" -o args | sed 1d)
+              DIST_OS=`uname -s | tr [:upper:] [:lower:] | tr -d [:blank:]`
+
+              case "$DIST_OS" in
+                'hp-ux')PID_COMMAND=$(UNIX95= ps -p "${ROOT_INSTANCE_PID}" -o args | sed 1d);;
+                *)PID_COMMAND=$(ps -p "${ROOT_INSTANCE_PID}" -o args | sed 1d);;
+              esac
               if [ "${PID_COMMAND#*$MAIN}" != "$PID_COMMAND" ]; then
```
gnodet: Would you mind leveraging and extending the `detectos` function instead ? https://github.com/apache/karaf/blob/master/assemblies/features/base/src/main/filtered-resources/resources/bin/inc#L28-L57

baptistedaroit: As requested, we updated the `detectos` function to add a specific variable to be associated with the ps command on HP-UX platforms when the `-o` option is used.

This was not mentioned on the [KARAF-5753](https://issues.apache.org/jira/browse/KARAF-5753) but we also provide an improvement for the following line in shell scripts:

 `if [ -z $READLINK_EXISTS ]; then`

**$READLINK_EXISTS** was not protected, causing error message on some platforms where the variable was empty (see below):

`client[2]: test: Specify a parameter with this command.`
`client[17]: readlink:  not found.`

assemblies/features/base/src/main/filtered-resources/resources/bin/inc
```diff
@@ -44,16 +45,20 @@ detectOS() {
...
+            export LDR_CNTRL=MAXDATA=0xB0000000@DSA
+            echo ${LDR_CNTRL}
             ;;
         OS400*)
             os400=true
             ;;
+        HP-UX*)
+            hpux=true
+            # For HP-UX, set an environment variable
+	    export PS_PREFIX="UNIX95= "
```
gnodet: Could you just fix the indentation with spaces and remove the leading tab please ?

baptistedaroit: OK, done.

baptistedaroit:
Tested on HP-UX B.11.31 ia64 with HP JDK 8 Java(TM) SE Runtime Environment (build 1.8.0.12-hp-ux-b1)

PR #511, opened 2018-05-22
[KARAF-5693] Run the processing synchronously after downloads

gnodet:


gnodet:
Retest this please

karendolan:
@gnodet Does this synchronize both the downloading of the libraries and  also the downloaded() shared activity that happens after?

karendolan:
In other words, will this change slow down library downloads?

gnodet:
No, it only synchronizes the shared activity and execute them in the order the libraries were called

karendolan:
That's great! Thanks.

jbonofre:
retest this please

jbonofre:
This PR actually breaks resolver verify due to concurrent modification in in the await method (on the callback hashmap). I tried to use a `ConcurrentHashMap` in this time the resolver behavior is not correct. I will take a deeper look.

PR #513, opened 2018-05-23
[KARAF-5765] Enhance karaf-service on HP-UX 

baptistedaroit:
Add display of HP-UX service instructions + fix bad usage of shell commands in karaf-service script when used on HPUX platform


jbonofre:
retest this please

PR #518, opened 2018-05-25
[KARAF-5737] Upgrade to Aries JPA 2.7.0

jbonofre:


jbonofre:
I have to fix the `enterprise` feature verify.

PR #531, opened 2018-06-18
[KARAF-5787] Improve scheduler whiteboard handler to allow users to define scheduler service properties in raw String

jbonofre:


jbonofre:
retest this please

PR #534, opened 2018-06-28
Update surefire + apache parent versions

coheigea:
Update the Surefire plugin (required for Java 10) + update the apache parent version


jbonofre:
Last time that I updated to this surefire version, it broke the itests and surefire (as described in the corresponding jira). I think it would require a new pax exam version supporting this surefire version.

coheigea:
The build and tests all worked fine for me (with Java 8)?

PR #540, opened 2018-07-05
[KARAF-5805] Add feature required item field to JmxFeature CompositeData

jamesnetherton:
https://issues.apache.org/jira/browse/KARAF-5805


jamesnetherton:
Must we build with SNAPSHOTs? Sorta screws CI up when they're not available in any remote repo.

PR #543, opened 2018-07-10
[KARAF-5809] fallback for 'simple' host.key

jutzig:
if the KeyPairProvider encounters a non-valid PEM it tries an
ObjectInputStream as a fallback (the old 'simple' format).
If that succeeds, it rewrites the keypair into PEM format


jbonofre:
retest this please

PR #549, opened 2018-07-12
[KARAF-5814] Upgrade to Aries Blueprint Core 1.10.0 and Blueprint CM 1.3.0

jbonofre:


jbonofre:
In order to avoid compatibility issue with Apache  Camel, I'm postponing this update.

PR #553, opened 2018-07-12
[KARAF-5810] No config file installed in custom distribution. (On master)

vanSabben:
Now config files also installed for custom distributions.

### DIFF
--- a/profile/src/main/java/org/apache/karaf/profile/assembly/Builder.java
+++ b/profile/src/main/java/org/apache/karaf/profile/assembly/Builder.java
@@ -1655,7 +1655,7 @@ public class Builder {
                 }
             }
 
-            new ConfigInstaller(etcDirectory, pidsToExtract)
+            new ConfigInstaller(homeDirectory, etcDirectory, pidsToExtract)
                 .installConfigs(feature, downloader, installer);
             // Install libraries
             List<String> libraries = new ArrayList<>();
--- a/profile/src/main/java/org/apache/karaf/profile/assembly/ConfigInstaller.java
+++ b/profile/src/main/java/org/apache/karaf/profile/assembly/ConfigInstaller.java
@@ -42,11 +42,11 @@ import org.slf4j.LoggerFactory;
 public class ConfigInstaller {
     private static final Logger LOGGER = LoggerFactory.getLogger(ConfigInstaller.class);
     private Path etcDirectory;
-    private Path homeDirectory;
+    private Path baseDirectory;
     private List<String> pidsToExtract;
 
-    public ConfigInstaller(Path homeDirectory, Path etcDirectory, List<String> pidsToExtract) {
-        this.homeDirectory = homeDirectory;
+    public ConfigInstaller(Path baseDirectory, Path etcDirectory, List<String> pidsToExtract) {
+        this.baseDirectory = baseDirectory;
         this.etcDirectory = etcDirectory;
         this.pidsToExtract = pidsToExtract;
     }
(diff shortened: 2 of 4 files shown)


profile/src/main/java/org/apache/karaf/profile/assembly/Builder.java
```diff
@@ -1655,7 +1655,7 @@ private void installStage(Profile installedProfile, Set<Feature> allBootFeatures
                 }
             }
 
-            new ConfigInstaller(etcDirectory, pidsToExtract)
+            new ConfigInstaller(homeDirectory, etcDirectory, pidsToExtract)
```
jbonofre: Why do we need the `homeDirectory` here ? Config should go in `etc`. And it seems to  be the `baseDirectory` (not home).

profile/src/main/java/org/apache/karaf/profile/assembly/ConfigInstaller.java
```diff
@@ -38,9 +42,11 @@
 public class ConfigInstaller {
     private static final Logger LOGGER = LoggerFactory.getLogger(ConfigInstaller.class);
     private Path etcDirectory;
+    private Path homeDirectory;
     private List<String> pidsToExtract;
-    
-    public ConfigInstaller(Path etcDirectory, List<String> pidsToExtract) {
+
+    public ConfigInstaller(Path homeDirectory, Path etcDirectory, List<String> pidsToExtract) {
+        this.homeDirectory = homeDirectory;
```
jbonofre: Same,  I don't think `homeDirectory` is required here.

profile/src/main/java/org/apache/karaf/profile/assembly/ConfigInstaller.java
```diff
@@ -93,7 +99,54 @@ public void installConfigs(Feature feature, Downloader downloader, ArtifactInsta
...
             }
+            for (ConfigFile configFile : content.getConfigfile()) {
+                if (pidMatching(FilenameUtils.getBaseName(configFile.getFinalname()))) {
+                    installConfig(downloader, configFile);
+                }
+            }
+        }
+    }
+
+    private void installConfig(Downloader downloader, ConfigFile pConfigFile) throws Exception {
```
jbonofre: The method should be named `installConfigFile` to  avoid confusion.

profile/src/main/java/org/apache/karaf/profile/assembly/ConfigInstaller.java
```diff
@@ -93,7 +99,54 @@ public void installConfigs(Feature feature, Downloader downloader, ArtifactInsta
...
+        String path = pConfigFile.getFinalname();
+        if (path.startsWith("/")) {
+            path = path.substring(1);
+        }
+
+        Path configFileTarget = homeDirectory.resolve(substFinalName(path));
```
jbonofre: I don't think it's a good idea to do this here. We should just copy the config file in the `etc` directory "raw".

profile/src/main/java/org/apache/karaf/profile/assembly/ConfigInstaller.java
```diff
@@ -93,7 +99,54 @@ public void installConfigs(Feature feature, Downloader downloader, ArtifactInsta
...
+
+        downloader.download(location, provider -> {
+            synchronized (provider) {
+                Files.createDirectories(configFileTarget.getParent());
+                Files.copy(provider.getFile().toPath(), configFileTarget, StandardCopyOption.REPLACE_EXISTING);
+            }
+        });
+    }
+
+    private String substFinalName(String finalname) {
```
jbonofre: Useless (at least not in the target of this class).

profile/src/main/java/org/apache/karaf/profile/assembly/Builder.java
vanSabben: The [documentation](https://karaf.apache.org/manual/latest/#_feature_configuration_files) says it is relative.
> If not specified, the location is relative from the KARAF_BASE variable. It’s also possible to use variable like ${karaf.home}, ${karaf.base}, ${karaf.etc}, or even system properties.

Each location can occur or it is relative to KARAF_BASE. KARAF_HOME is indeed not good. But is home not always equal to base in the AssemblyMojo? Here are no other instances possible? The builder name it home, but it is the custom distribution directory in the maven target directory. What do you think is best?

profile/src/main/java/org/apache/karaf/profile/assembly/ConfigInstaller.java
vanSabben: True, but baseDirectory is required.

profile/src/main/java/org/apache/karaf/profile/assembly/ConfigInstaller.java
vanSabben: Correct.

profile/src/main/java/org/apache/karaf/profile/assembly/ConfigInstaller.java
vanSabben: What do you mean by 'raw'? Do not use variables and the like? See comment / documentation above. It can be absolute or relative and use variables.

profile/src/main/java/org/apache/karaf/profile/assembly/ConfigInstaller.java
vanSabben: See comment / documentation above. It can be absolute or relative and use variables.

jbonofre:
See my comment in Jira.

PR #554, opened 2018-07-12
[KARAF-5810] No config file installed in custom distribution. (On karaf-4.1.x)

vanSabben:
Now config files also installed for custom distributions.


jbonofre:
See my comment in Jira.

PR #599, opened 2018-08-20
KARAF-5877 log4j2 appending error when using log4j2 XML configuration

bakejeyner:
See [this JIRA ticket|https://issues.apache.org/jira/browse/KARAF-5877] for more background information.

In short, `LogServiceLog4j2XmlImpl` will throw an `org.w3c.dom.DOMException: NOT_FOUND_ERR` when the corresponding `log4j2` configuration file has an element with children as the last element in the `<Loggers>` element. This causes the `log:tail` command to display the logs and error out immediately.


bakejeyner:
Closed because the changes I was making ended up being larger in scale than I first thought.

See https://issues.apache.org/jira/browse/KARAF-5877 to continue progress.

PR #602, opened 2018-09-05
[KARAF-5887] Fix NPE on ConsoleSessionImpl

NicolasOttaviani:
Fix NullPointerException when starting shell using bin/shell
See Jira: https://issues.apache.org/jira/browse/KARAF-5887

### DIFF
--- a/shell/core/src/main/java/org/apache/karaf/shell/impl/console/ConsoleSessionImpl.java
+++ b/shell/core/src/main/java/org/apache/karaf/shell/impl/console/ConsoleSessionImpl.java
@@ -135,7 +135,7 @@ public class ConsoleSessionImpl implements Session {
         }
 
         // Create session
-        if (in == null) {
+        if (in == null || out == null || err == null) {
             session = processor.createSession(
                 jlineTerminal.input(),
                 jlineTerminal.output(),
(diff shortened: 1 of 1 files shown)


shell/core/src/main/java/org/apache/karaf/shell/impl/console/ConsoleSessionImpl.java
```diff
@@ -135,7 +135,15 @@ public ConsoleSessionImpl(SessionFactory factory,
         }
 
         // Create session
-        session = processor.createSession(in, out, err);
+        if (in == null) {
```
jbonofre: Just in case, it's maybe better to test `out` and `err` as well:

```
if (in == null || out == null || err == null) {
```

jbonofre:
Can you please rebase with `master` ? I fixed the build with #606

shell/core/src/main/java/org/apache/karaf/shell/impl/console/ConsoleSessionImpl.java
NicolasOttaviani: Yes, sorry for the mistake, i have push a new commit

PR #604, opened 2018-09-07
Updated Scheduler feature to support persistence

miroslav-beranic:
Updated Scheduler feature to support persistence ( Oracle JDBC, C3P0 JDBC pool, Serialization ).

- ScheduleOptions was made Serializable, with addition to move Trigger creation from InternalScheduleOptions's constructor to method compile().
- getJobs() method from Scheduler interface returns String names and not Objects
- Added custom StdScheduler and StdSchedulerFactory classes, to implemented changes
- Added SchedulerStorage as support for "backward compatibility" and "short lived" jobs. This is not good place to use it.
- Updated QuartzJobExecutor to remove non-serializable items from JobDataMap storage
- Updated Scheduler Shell commands to use changed Scheduler API ( job names are as keys )

### DIFF
--- a/scheduler/pom.xml
+++ b/scheduler/pom.xml
@@ -64,11 +64,11 @@
                             org.apache.karaf.scheduler;version=${project.version};-noimport:=true
                         </Export-Package>
                         <Import-Package>
-                            !com.mchange.*,
-                            !oracle.*,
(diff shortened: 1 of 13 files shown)


miroslav-beranic:
Need to check, why build failed.

PR #638, opened 2018-10-24
 Updated Col used by ShellTable. Added support for changing color of a specific cell in a column based on the cell content.

labero:


labero:
Seems to be fixed now. To be honest, I had some problem finding the error message from the javadoc generation.

labero:
I realized that karaf 4.2 is not supporting lower java versions that 8. So I rebased, squashed some commits, removed the IColorProvider interface and replaced it with the standard "Function" interface from java 8.

However, the Jenkins failure (JMXSecurityTest). I can't see how that is affected by my changes. Is there anything I can do to resolve it?

jbonofre:
The build issue is not related, it's a flaky test that I will fix.

jbonofre:
retest this please

PR #647, opened 2018-11-06
Update XBean

coheigea:


jbonofre:
There's a Jira about that and a PR coming sync with ASM 7 update as well.

PR #648, opened 2018-11-06
[KARAF-5984] Add spotless maven plugin support

jbonofre:


jbonofre:
retest this please

jbonofre:
We didn't have a strong consensus about this change. So I cancel this change for now.

PR #670, opened 2018-12-03
[ARIES-1793] Upgrade to blueprint-core 1.10.0

Smasherr:


cschneider:
Ok. I will revert. The issue with jpa is important though and actually not having interceptors in blueprint will cause big problems in many applications. So lets release a blueprint core 1.9.1 with just the interceptor fix but not the change that breaks camel. Can you point me to what broke camel?

PR #672, opened 2018-12-03
[KARAF-6015] Upgrade to Aries SpiFly 1.2

jbonofre:


graben:
```
<bundle dependency="true" start-level="30">mvn:org.ow2.asm/asm-all/${asm.version}</bundle>
<bundle dependency="true" start-level="30">mvn:org.apache.aries/org.apache.aries.util/${aries.util.version}</bundle>
```
Those lines can be removed too from jetty feature in standard feature file. ASM is embedded an aries util has been replaced by bnd impl.

jbonofre:
retest this please

PR #691, opened 2018-12-09
[KARAF-6042] Fix flaky WebSocketExampleTest (if the http service takes time to startup)

jbonofre:


jbonofre:
retest this please

PR #708, opened 2018-12-30
Wrap NIO IOException where it's thrown.

peculater:
Hi there.  Over at the Apache Tamaya project, we're using the karaf-maven-plugin on our extensions.  The extensions themselves are built as a bunch of modules in a maven pom, and some have dependencies between them.  When karaf-maven-plugin resolves one of those dependencies, it will often resolve it to the target/classes directory of the newly-built project, and not to a jar file which hasn't yet been installed.  Importantly that means that any operations that assume it is a jar will fail. That's been going on for a while and is probably fine.

Starting with karaf v4.1.3 and continuing through 4.2.2 we started seeing the surprising error message `Unable to create features.xml file: java.io.IOException: Is a directory` when running maven.  Turns out the IOException from the jar operations is no longer being caught and converted to a warning.  I found http://mail.openjdk.java.net/pipermail/nio-dev/2014-December/002877.html which taught me that while old-style `FileInputStream` throws an exception if you try to open a stream on a directory, NIO is happy to create the stream but throws the exception once you read from it. https://github.com/peculater/karaf/commit/ed797960d2a43eee8c04a9572478320af2420cfb added the NIO logic but didn't move the catch for the sneaky exception.

### DIFF
--- a/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/features/GenerateDescriptorMojo.java
+++ b/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/features/GenerateDescriptorMojo.java
@@ -667,7 +667,7 @@ public class GenerateDescriptorMojo extends MojoSupport {
      * Extract the MANIFEST from the give file.
      */
 
-    private Manifest getManifest(File file) throws IOException {
+    private Manifest getManifest(File file) {
         final InputStream is;
         try {
             is = Files.newInputStream(file.toPath());
@@ -686,6 +686,9 @@ public class GenerateDescriptorMojo extends MojoSupport {
                 }
                 return m;
             }
+        } catch (IOException e) {
+            getLog().warn("Error while reading artifact", e);
+            return null;
         }
     }
 


tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/features/GenerateDescriptorMojo.java
```diff
@@ -686,6 +686,9 @@ private Manifest getManifest(File file) throws IOException {
                 }
                 return m;
             }
+        } catch (IOException e) {
+            getLog().warn("Error while reading artifact", e);
+            return null;
```
jbonofre: I would prefer to keep `throws IOException` and add log. Something like;

```
private Manifest getManifest(File file) throws IOException {
  final InputStream is;
  try {
    is = Files.newInputStream(file.toPath());
    ...
  } catch (IOException e) {
    getLog().warn("Error while reading artifact", e);
    throw e;
 }
```

peculater: Unfortunately that doesn't solve the problem I'm having.  That same exception in v4.1.2 would have been turned into a warning at https://github.com/apache/karaf/pull/708/files#diff-0a5a9e3c3d44c1b639a942a084b31627L676 and not re-thrown.  I'm really trying to get back to that behaviour.

jbonofre: I don't understand: my proposal is to both log AND rethrow. So it should work as you expect (as you expect a warn).

peculater: Yeah, whether the warning happens or not, I need the "Is a directory" IOException to not show up outside of this method.   Previously the exception would be thrown on opening the FileInputStream, so the exception would get caught at line 674 and the method would return null.  Now (after converting to NIO) the exception is thrown when the stream is read from, so it is thrown on line 682 and is never caught.  That's why I'm adding the catch block where I am and returning null.
I hope that's more clear?

PR #718, opened 2019-01-15
Consolidate LoginModule functionality in the abstract class. Also imp…

coheigea:
…lement abort() properly. In addition, the commit logic in GSSAPILdapLoginModule was not correct.


jbonofre:
retest this please

PR #728, opened 2019-01-17
[KARAF-6093] Fix jax-ws packages being exported by bundle 0

brjeter:
Removing `javax.xml.soap` from `jre.properties` wasn't enough (https://github.com/apache/karaf/pull/719). Also needed to remove `javax.xml.ws` packages since they use `javax.xml.soap`.

### DIFF
--- a/assemblies/features/base/src/main/filtered-resources/resources/etc/jre.properties
+++ b/assemblies/features/base/src/main/filtered-resources/resources/etc/jre.properties
@@ -445,14 +445,6 @@ jre-1.8= \
  javax.xml.transform.stax, \
  javax.xml.transform.stream, \
  javax.xml.validation, \
- javax.xml.ws;version="2.2", \
- javax.xml.ws.handler;version="2.2", \
- javax.xml.ws.handler.soap;version="2.2", \
- javax.xml.ws.http;version="2.2", \
- javax.xml.ws.soap;version="2.2", \
- javax.xml.ws.spi;version="2.2", \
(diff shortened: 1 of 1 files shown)


jbonofre:
retest this please

jbonofre:
This change breaks CXF by default (at least the Karaf examples). I will take a look.

jbonofre:
retest this please

brjeter:
Sounds good. Let me know if I can help. The `cxf-specs` feature should be taking care of exporting these packages in the container when they're not provided by the container itself.

ponziani:
Any news on this ? I believe it would fix the issue described at http://karaf.922171.n3.nabble.com/Loader-constraint-violation-javax-xml-soap-SOAPFault-td4056103.html as well.

jbonofre:
@ponziani as explained on the mailing list, there's a new set of spec features coming. I will fix this one and related. It will be included in Karaf 4.3.0.RC2 and 4.2.9. Thanks for your patience.

jbonofre:
This will be part on a more global PR aka "features specs".

ponziani:
@jbonofre, we're still having the same problem in Karaf 4.3.2. Looking at the source code, it looks like the `javax.xml.ws` packages were not actually removed from jre.properties. 

Can you please point me to information about the more gloabl PR aka "feature specs" ?

I've temporarily patched our custom distribution with a custom jre.properties, but I would like to see a better solution.

PR #729, opened 2019-01-18
[KARAF-6089] Prevent deadlock on bundle lock between FelixStartLevel …

grgrzybek:
…and features threads


grgrzybek:
don't merge yet - I have better fix

PR #732, opened 2019-01-19
[KARAF-6058] Fix default etc/keys.properties and update documentation about ssh key authentication

jbonofre:


jbonofre:
retest this please

PR #738, opened 2019-01-22
KARAF-6109: Do filter effective installed features for custom distributions for non full bundle-report

graben:
@grgrzybek : Plz have a look, maybe we get this into 4.2.3!

fyi @jbonofre

### DIFF
--- a/profile/src/main/java/org/apache/karaf/profile/assembly/Builder.java
+++ b/profile/src/main/java/org/apache/karaf/profile/assembly/Builder.java
@@ -317,6 +317,7 @@ public class Builder {
     private FeaturesProcessing featuresProcessing = new FeaturesProcessing();
     private Map<String, String> translatedUrls;
     private Blacklist blacklist;
+    private String generatedBootFeatureName;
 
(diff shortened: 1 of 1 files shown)


grgrzybek:
is this the same patch you've sent in email (topic: "Version alignment report for Karaf 4.2.3-SNAPSHOT - 1/2")?

graben:
@grgrzybek : It's a little bit beautified.

grgrzybek:
Thanks. My intention for _non full bundle-report_ was to exclude blacklisted bundles/features/repositories.
I'll check if there's a need for 3rd kind of report so we have 3:
1. all
2. all but blacklisted
3. only installed

jbonofre:
I think that report could be full and the user/dev can find in there (by section).

grgrzybek:
Good idea. I can (try) add a switch to select kind of report within single XML.

grgrzybek:
@graben @gnodet here's my solution ;)
http://people.apache.org/~ggrzybek/bundle-report-karaf-new.xml

graben:
@grgrzybek : Looks good 👍

PR #608, opened 2018-09-10
Updated Scheduler feature to support persistence

miroslav-beranic:
Updated Scheduler feature to support persistence

Updated Scheduler feature to support persistence ( Oracle JDBC, C3P0 JDBC pool, Serialization ). Updated import packages resolution.

- ScheduleOptions was made Serializable, with addition to move Trigger creation from InternalScheduleOptions's constructor to method compile().
- getJobs() method from Scheduler interface returns String names and not Objects
- Added custom OSGi StdScheduler and StdSchedulerFactory classes, to implemented changes
- Added SchedulerStorage as support for "backward compatibility" and "short lived" jobs. This is not good place to over-use it.
- Updated QuartzJobExecutor to remove non-serializable items from JobDataMap storage
- Updated Scheduler Shell commands to use changed Scheduler API ( job names are as keys - String )
- Made SchedulerError extend RuntimeException, to have cleaner code.
- All bundle import dependencies were made optional

``` 
$ bundle:install -l 30 -s mvn:com.oracle/jdbc-osgi/12.1.0.1

$ feature:install jndi transaction aries-blueprint jdbc scr pax-jdbc-oracle

$ bundle:install -l 30 -s mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.quartz/2.2.4-SNAPSHOT ( Custom build to include Quartz 2.2.1 )

$ bundle:install -l 30 -s mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.c3p0/0.9.5.2_3-SNAPSHOT ( Compatible version with Quartz 2.2.1 )

$ feature:install scheduler

$ scheduler:list

$ feature:repo-add mvn:org.apache.karaf.examples/karaf-scheduler-example-features/4.2.2-SNAPSHOT/xml
$ feature:install karaf-scheduler-example
```

PR #746, opened 2019-01-31
[KARAF-6123] Add scheduler storage feature allowing to implement different storage

jbonofre:
Credit To: Miroslav Beranic <miroslav@mibesis.si>


jbonofre:
This PR overrides #608

PR #765, opened 2019-02-13
[KARAF-6156] Update JaasHelper to support extending Karaf JAAS (4.1.x)

mattrpav:
Use an .isAssignableFrom() instead of class name string comparison. This allows for extending Karaf JAAS [User,Group,Role]Principal classes

### DIFF
--- a/util/src/main/java/org/apache/karaf/util/jaas/JaasHelper.java
+++ b/util/src/main/java/org/apache/karaf/util/jaas/JaasHelper.java
@@ -61,15 +61,22 @@ public class JaasHelper {
         if (index > 0) {
             clazz = requestedRole.substring(0, index);
             role = requestedRole.substring(index + 1);
+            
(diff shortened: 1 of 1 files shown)


jbonofre:
Please, just open PR based on `master`. The automatic Jenkins job is based on `master`. Thanks anyway ;)

mattrpav:
Do you no longer want to receive back-ported PRs?

jbonofre:
It's up to you, I just mean that the build could fail on Jenkins as Jenkins PR build is only based on master.

jbonofre:
retest this please

mattrpav:
> It's up to you, I just mean that the build could fail on Jenkins as Jenkins PR build is only based on master.

Its easy enough for me to make back-ported PRs.. if you don't mind the build break noise, I'll keep them coming as long as they are helpful.  Won't hurt my feelings either way.

PR #772, opened 2019-03-06
[KARAF-6183] Fix override for classified artifacts.

Smasherr:
We take now the pattern from override.properties, which
matches best (i.e. has the biggest length), not the last one


grgrzybek:
:+1:

PR #789, opened 2019-03-21
Add docker related MOJOs and corresponding example

jbonofre:


fpapon:
AWESOME!

tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/DockerMojo.java
```diff
@@ -0,0 +1,64 @@
...
+    private File location;
+
+    @Parameter(defaultValue = "karaf")
+    private String imageName;
+
+    @Override
+    public void execute() throws MojoExecutionException, MojoFailureException {
+        getLog().info("Creating Docker image");
+        try {
+            Process process = new ProcessBuilderFactoryImpl().newBuilder()
```
rmannibucau: Tink you need to optionally log in before ;)
Also it can be worth using jib to control the layers properly (karaf, deps, the app, config likely) and it will not require the daemon to be present (which avoids version issues ;))

That said it is an awesome move! +1

jbonofre: Thanks for the update, I wanted to provide something very simple. I fully agree about jib, I will add a profile to illustrate this.

Thanks !!!!

jbonofre:
Split into two PRs.

PR #791, opened 2019-03-23
KARAF-6207, bootFeatures sometimes being ignored

lkiesow:
According to the documentation, the maven-karaf-plugin will install
features listed in the installedFeatures in the "system" internal
repository. Features listed in bootFeatures are additionally added to
the boot-features in the features service configuration file.

Hence, the configuration options do not conflict and a configuration
like this should result in the features being installed and added to the
boot-features configuration:

    <configuration>
      <bootFeatures>
        <feature>foo</feature>
      </bootFeatures>
      <installedFeatures>
        <feature>foo</feature>
      </installedFeatures>
    </configuration>

In fact, the configuration should be equivalent to `foo` being listed
just as a bootFeature.

But what actually happens is that the installedFeature will overwrite
the bootFeature and the feature is not added to the boot-features
configuration.

This patch switches the handling of bootFeatures and installedFeatures
which makes a feature configured in both sections a bootFeature,
adhering to the specifications of both configurations.

### DIFF
--- a/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/AssemblyMojo.java
+++ b/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/AssemblyMojo.java
@@ -536,14 +536,6 @@ public class AssemblyMojo extends MojoSupport {
                .bundles(toArray(startupBundles))
                .profiles(toArray(startupProfiles));
 
-        // Boot stage
-        builder.defaultStage(Builder.Stage.Boot)
-                .kars(toArray(bootKars))
-                .repositories(bootFeatures.isEmpty() && bootProfiles.isEmpty() && installAllFeaturesByDefault, toArray(bootRepositories))
-                .features(toArray(bootFeatures))
-                .bundles(toArray(bootBundles))
-                .profiles(toArray(bootProfiles));
-
         // Installed stage
         builder.defaultStage(Builder.Stage.Installed)
                 .kars(toArray(installedKars))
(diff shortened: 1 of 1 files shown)


jbonofre:
It's one or the other: it doesn't makes sense to have the `foo` feature in both `installed` and `boot` features. If you put in `boot` feature, it's implicitly in the `system` folder as well.

jbonofre:
I consider that you configuration is not correct as it's "exclusive": the feature should be ONLY in boot, installed or startup.

lkiesow:
Hi @jbonofre,
thanks for taking a look at this.

The configuration makes sense for a use case where you have a parent pom specifying all features which should be installed and then different child poms specifying specific features to boot up (but leaving the option to start the others):

```
parent
 |   installedFeatures: foo, bar, …
 +-child1
 |     bootFeatures: foo, …
 +-child2
 |     bootFeatures: bar, …
```

That makes sense because [the documentation I found](https://svn.apache.org/repos/asf/karaf/site/production/manual/latest/custom-distribution.html#_plugin_configuration) specified a bootFeature as an installedFeature which also gets an entry in the boot-features configuration. I didn't find any documentation of your claim that the feature configuration is meant to be exclusive.

In any case, if you decide to make this exclusive, can you document that and (probably more important) make the maven-karaf-plugin throw an error and fail in such a case?  Because right now the feature configurations are also not exclusive in the code. In the code, the behavior is that the installedFeature configuration takes precedence over the bootFeature configuration.

jbonofre:
Yes, that's an implicit assertion. We should check that a feature doesn't appear in installed, boot, startup (it's only one). I will do the change.

jbonofre:
Waiting for the "exclusive" check, I'm merging your PR.

PR #794, opened 2019-03-24
KARAF-6209: Add SortedProperties class to karaf tooling

graben:
@jbonofre : Could you please have a look on this? Merci!


jbonofre:
Can you please provide an abstract for this change (use case) ? Thanks !

jbonofre:
OK, it's for the readability of the properties. Fair enough ;)

PR #802, opened 2019-04-01
KARAF-6220, add principal info to file and log audit logs

sjhiggs:


grgrzybek:
Looks good, thanks @sjhiggs. I'd only complain about:
```java
if(principalInfos.size() > 0) {
```
(lack of space between `if` and `(`)

sjhiggs:
thanks @grgrzybek, fixed.

jaas/modules/src/test/java/org/apache/karaf/jaas/modules/audit/AbstractAuditLoginModuleTest.java
```diff
@@ -0,0 +1,37 @@
+package org.apache.karaf.jaas.modules.audit;
```
jbonofre: Please, add the ASF header.

sjhiggs: done

PR #810, opened 2019-04-12
KARAF-6235:use System.currentTimeMillis()

bd2019us:
new Date() is just a thin wrapper around System.currentTimeMillis(). Using System.currentTimeMillis() can help speed up the system


bd2019us:
> Can you please amend the commit to use `[KARAF-6235] ...." commit message ?

Hello.     I have modified the commit message.

PR #811, opened 2019-04-13
[KARAF-6229] Fix dual install/deploy of zip archive by the karaf-maven-plugin

jbonofre:


jbonofre:
retest this please

tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/ArchiveMojo.java
```diff
@@ -116,39 +116,44 @@ public void execute() throws MojoExecutionException, MojoFailureException {
         org.apache.maven.artifact.Artifact artifact = project.getArtifact();
         artifact.setFile(targetFile);
         try {
-            //now pack up the server.
-            if(archiveTarGz){
-                archive("tar.gz");
+            if (project.getPackaging().equals("karaf-assembly") && !archiveTarGz && !archiveZip) {
+                throw new IllegalArgumentException("For karaf-assembly packaging, you have to specify at least one archive type (tar.gz or zip)");
```
lkiesow: Unfortunately, this line breaks our build and a few of our use cases where we do not pack our assemblies.

For example, we have a development assembly which we have end up in a special folder unpacked and prepared to be launched. We could, of course, pack and then automatically unpack it but that would just be a waste of time.

We also have some assemblies which we pack with additional files (documentation, …) which is handled separately. Again a use-case for assemblies unpacked by the karaf-maven-plugin.

@jbonofre, would you mind if I create a patch removing this check again? So that creating, but not packing assemblies works again.

jbonofre: Can you share a simple `pom.xml` ? I guess you want to use `assembly` (not `archive`) right ?

lkiesow: This is a simplified version of one of the cases were we disabled both `archiveTarGz` and `archiveZip`:

https://github.com/lkiesow/opencast/blob/karaf-425-example/assemblies/dist-develop/pom.xml#L92-L128

jbonofre: Ok, you just use `karaf-assembly` packaging and no `executions`. I better understand now. I will fix that, sorry about that, I missed this use case ;)

lkiesow: No problem. Changes are expected to break things, aren't they ;-P

Pull request #833 fixes the problem by simply allowing to deactivate both archive types. That works for us and that's what we have done in the past.

Of course, I'm not opposed to any other solution if you have any. I've seen that some other plugins have flags to skip certain goals entirely but I never really had a close look at any maven plugin so I'm not sure what the best solution is.

PR #820, opened 2019-04-24
[KARAF-6254] Use backticks for command substitution

baptistedaroit:


jbonofre:
I'm testing your change on different Unix (AIX, several Linux distro, ...) to see if it works fine (I remember I had some issues with back quote with some shell).

PR #822, opened 2019-04-26
[KARAF-6258] Do not print error for user interrupted script

lkiesow:
Karaf allows for the usage of commands like `log:tail` in
`etc/shell.init.script` which is handy for example for a development
assembly. But if such a command is canceled by a user (e.g. by hitting
Ctrl+C) an error message is logged and printed to stderr, claiming an
error in the initialization script.

While keeping the debug logging, this patch suppresses the stderr
message on user interrupts.


lkiesow:
Sure, no problem. Done.

PR #825, opened 2019-05-02
Updating Jetty

coheigea:
Updating Jetty to avoid a bunch of CVEs.


jbonofre:
That's not fully correct as it requires update on pax web. I'm closing this pr toi create another one including pax web.

PR #837, opened 2019-05-14
KARAF-6276: Delete update file in finally block

htreu:
As described in [KARAF-6276](https://issues.apache.org/jira/browse/KARAF-6276) an intermediate bundle update file might not be deleted when an exception occurs during update.
This should fix this by moving the `file#delete` call into a finally block.

Signed-off-by: Henning Treu <henning.treu@instana.com>

### DIFF
--- a/features/core/src/main/java/org/apache/karaf/features/internal/service/BundleInstallSupportImpl.java
+++ b/features/core/src/main/java/org/apache/karaf/features/internal/service/BundleInstallSupportImpl.java
@@ -140,13 +140,17 @@ public class BundleInstallSupportImpl implements BundleInstallSupport {
 
     @Override
     public void updateBundle(Bundle bundle, String uri, InputStream is) throws BundleException {
+        File file = null;
         // We need to wrap the bundle to insert a Bundle-UpdateLocation header
         try {
-            File file = BundleUtils.fixBundleWithUpdateLocation(is, uri);
+            file = BundleUtils.fixBundleWithUpdateLocation(is, uri);
             bundle.update(new FileInputStream(file));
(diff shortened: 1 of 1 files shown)


features/core/src/main/java/org/apache/karaf/features/internal/service/BundleInstallSupportImpl.java
```diff
@@ -140,13 +140,17 @@ public Bundle installBundle(String region, String uri, InputStream is) throws Bu
...
         // We need to wrap the bundle to insert a Bundle-UpdateLocation header
         try {
-            File file = BundleUtils.fixBundleWithUpdateLocation(is, uri);
+            file = BundleUtils.fixBundleWithUpdateLocation(is, uri);
             bundle.update(new FileInputStream(file));
-            file.delete();
         } catch (IOException e) {
             throw new BundleException("Unable to update bundle", e);
+        } finally {
+            if (file != null && file.exists()) {
```
skitt: Just a minor comment while the checks run — I don’t think there’s much point in calling `.exists()` before `.delete()` (there’s still a TOCTTOU race, and the number of syscalls isn’t reduced).

skitt:
> By the `exists()` test is not required but fair enough.

It was removed in the update ;-).

jbonofre:
Ah yes, I forgot to update my browser ;) So please, change the commit message.

htreu:
@jbonofre done.

htreu:
Some itest failed but I can not make much sense out of it. Seems to be unrelated. Maybe restart the build?

skitt:
retest this please

PR #842, opened 2019-05-15
Cleanup

skitt:
This patch series groups a number of largely mechanical clean-ups.


skitt:
@jbonofre do you want me to file an issue for these?

jbonofre:
@skitt no need of a Jira.

If you don't mind, I'm doing the review and then I will squash in one commit. Agree ?

skitt:
> If you don't mind, I'm doing the review and then I will squash in one commit. Agree ?

Yup, that’s fine. I pushed the individual commits so that each change would be easily identifiable and reviewable, but it would create a lot of noise in the history!

jbonofre:
retest this please

PR #849, opened 2019-05-21
[KARAF-2894] Add option to delete configurations/configuration files while uninstalling feature

jbonofre:


jbonofre:
I would like to add few tests/itests around that.

PR #850, opened 2019-05-21
Switch to using https in the poms

coheigea:


jbonofre:
I have PRs ready for Karaf subprojects.

PR #879, opened 2019-06-19
Maven plugin feature to define custom bundle location prefixes

ricardosilva88:
Allowing custom deployers on karaf 4 plugin, this comes in hand of the following PR https://github.com/apache/karaf/pull/489 that was merged in the past for karaf 3.

The idea is to have the following definition in karaf-maven-plugin
`<extraProtocols>
         <param>pentaho-webjars</param>
            <param>pentaho-platform-plugin-mvn</param>
          </extraProtocols>
`

which allows the following declaration
`<bundle>pentaho-webjars:mvn:org.webjars.npm/underscore/1.8.3</bundle>`

Please note, that we need this because we are doing our own karaf assembly.

### DIFF
--- a/features/core/src/main/java/org/apache/karaf/features/internal/download/impl/DownloadManagerHelper.java
+++ b/features/core/src/main/java/org/apache/karaf/features/internal/download/impl/DownloadManagerHelper.java
@@ -16,15 +16,35 @@
  */
 package org.apache.karaf.features.internal.download.impl;
 
+import java.util.Collection;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 public final class DownloadManagerHelper {
 
-    private static final Pattern IGNORED_PROTOCOL_PATTERN = Pattern.compile("^(jar|war|war-i|warref|webbundle|wrap|spring|blueprint):.*$");
+    private static final String DEFAULT_IGNORED_PROTOCOL_PATTERN = "jar|war|war-i|warref|webbundle|wrap|spring|blueprint";
+    private static Pattern ignoredProtocolPattern;
+
+    static {
+        setIgnoredProtocolPattern(DEFAULT_IGNORED_PROTOCOL_PATTERN);
+    }
 
     private DownloadManagerHelper() {
         //Utility Class
+    }
+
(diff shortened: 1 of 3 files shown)


ricardosilva88:
requesting @jbonofre review

features/core/src/main/java/org/apache/karaf/features/internal/download/impl/DownloadManagerHelper.java
```diff
@@ -16,17 +16,37 @@
...
 package org.apache.karaf.features.internal.download.impl;
 
+import java.util.Collection;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 public final class DownloadManagerHelper {
 
-    private static final Pattern IGNORED_PROTOCOL_PATTERN = Pattern.compile("^(jar|war|war-i|warref|webbundle|wrap|spring|blueprint):.*$");
+    private static final String DEFAULT_IGNORED_PROTOCOL_PATTERN = "jar|war|war-i|warref|webbundle|wrap|spring|blueprint";
```
jbonofre: What's the rational to not use the pattern ? It should be equivalent right ?

jbonofre: Maybe it's worth to update the prefix pattern no ?

ricardosilva88: From my perspective, this pattern is the "default pattern", it will be changed only if there exists a <extraProtocols> tag, which <params> values will be added to this pattern, by reconstructing the pattern on run time.

ricardosilva88: It is allowing karaf-assembly-plugin users to define their own set of custom prefixes

jbonofre:
@ricardosilva88 do you mind to base your PR on `master` ? I will cherry-pick on `karaf-4.2.x` once merged on `master`.

ricardosilva88:
Since master and this branch have some differences, I will close this one and open another PR

PR #884, opened 2019-06-19
Remove SSH hostKeyFormat from the config file

coheigea:
It isn't used anywhere in the code.


coheigea:
@jbonofre Looks like it is already removed from there, I did a grep and can't find a reference to it anywhere else in the source.

PR #886, opened 2019-06-20
Adding support for SSH using elliptic curves

coheigea:
Right now you can't SSH into Karaf using elliptic curve keys. The fix involves creating a new config parameter, that defaults to etc/host.key.pub to write out the public key that's generated.

Tested successfully in Karaf using an OpenSSH client.


jbonofre:
retest this please

PR #889, opened 2019-06-21
Switch to use java.nio.file.Files.createTempFile

coheigea:
According to the Javadoc:

The Files.createTempFile method provides an alternative method to create an empty file in the temporary-file directory. Files created by that method may have more restrictive access permissions to files created by this method and so may be more suited to security-sensitive applications.


jbonofre:
retest this please

PR #896, opened 2019-07-01
KARAF-6350 - Add support for elliptic keys in the PublicKeyLoginModule

coheigea:
This involves flipping the logic in PublicKeyLoginModule. Instead of converting the client key to a String + comparing it to the String in keys.properties, it takes the String in keys.properties and converts it into a Public Key. It's easier doing this for elliptic curve keys, as the encoding rules are pretty complex. I also added a bunch of tests.


jbonofre:
retest this please

coheigea:
@jbonofre Done

jbonofre:
retest this please

PR #899, opened 2019-07-05
[KARAF-6351] Handle package-less classes

skitt:
The classes command unconditionnally tries to extract a package name,
which fails with classes which don't have one. This patch checks for
the presence of "/" before attempting to use it as a delimiter.
Classes with no package are assumed to not be exported.


jbonofre:
I didn't validate the review (sorry I'm in vacation this week).

skitt:
> I didn't validate the review (sorry I'm in vacation this week).

Ah, sorry, do you want me to revert?

jbonofre:
@skitt no, it's OK, no problem, I hope it's OK (I saw an issue in a previous PR/commit not reviewed who broke the password encryption, that's why it's important to have a four eyes validation).

PR #901, opened 2019-07-10
Removing default karaf.key

coheigea:
Karaf bundles a default "karaf.key" for SSH. This is not good security practice, and it doesn't really have any practical use unless the user uncomments the hard-coded key in etc/keys.properties. The latter is not something we should support, if the user wants to use keys.properties, then the user should generate a keypair and put the public key into keys.properties as explained in the manual.

I tested that SSH'ing into Karaf using both a password, and a key defined in keys.properties works correctly with this fix, both for openssh and using bin/client.


jbonofre:
retest this please

jbonofre:
It seems that the GitHub/Jenkins integration doesn't work since yesterday (not only for Karaf, but globally). I'm checking with infra.

jbonofre:
retest this please

jbonofre:
retest this please

jbonofre:
I gonna test this PR manually waiting fix on github/jenkins integration.

jbonofre:
retest this please

jbonofre:
retest this please

jbonofre:
retest this please

jbonofre:
The tests failures have been fixed. I'm rebasing to test.

PR #923, opened 2019-09-04
KARAF-6405 - Make sure Config files are jailed to the container

coheigea:


features/core/src/main/java/org/apache/karaf/features/internal/service/FeatureConfigInstaller.java
```diff
@@ -252,12 +252,16 @@ private void installConfigurationFile(String fileLocation, String finalname, boo
         finalname = substFinalName(finalname);
 
         File file = new File(finalname);
+        if (!file.getCanonicalPath().startsWith(System.getProperty("karaf.base"))) {
```
j3rem1e: Will that prevent me from installing a config under "karaf.etc" which is not under "karaf.base" ?

jbonofre: I'm -1 about this as some users have `karaf.etc` "outside" of `karaf.base`. I would replace with:

```
if (!file.getCanonicalPath().startsWith(System.getProperty("karaf.etc")))
```

jbonofre: Indeed, I think it's better to use `karaf.etc` rather than `karaf.base`.

jbonofre: Generally speaking, I don't want to be too much strict for users. Some users have "exploded" Karaf in several folders. This change is too restrictive to then IMHO.

coheigea: @jbonofre  I've fixed it to use karaf.etc now, please review again.

jbonofre:
retest this please

jbonofre:
retest this please

jbonofre:
retest this please

PR #936, opened 2019-09-16
[KARAF-6417] Make users properties file configurable (again) for Auto…

grgrzybek:
…EncryptionSupport


jbonofre:
@grgrzybek I guess the same should be applied to `master` right ?

grgrzybek:
sure - also to master.

PR #937, opened 2019-09-16
[KARAF-6418] Flag to mark transitive dependencies with dependency="true"

nantunes:
See https://issues.apache.org/jira/browse/KARAF-6418.

Adds a flag `markTransitiveAsDependency` indicating whether the plugin should mark transitive dependencies' bundle elements as a dependency. Defaults to `false` to keep the current behaviour.

For consistency, also adds the `markRuntimeScopeAsDependency` flag, indicating whether the plugin should also mark dependencies' with the runtime scope. Defaults to `true`.

Note: In fact I'm needing this in the 4.2.x branch.


jbonofre:
I like what you did: it's really clean including test. Congrats ! Let me take a complete deeper look. Thanks !

jbonofre:
The test failure is not related to your change, it's a flaky test we have in Karaf (I have to improve it ;)).

PR #944, opened 2019-09-20
[KARAF-6423] Upgrade to JAXB 2.3.3

jbonofre:


ffang:
the problem is that there's no glassfish jaxb 2.3.3 yet. And CXF also want it for another issue
https://github.com/eclipse-ee4j/jaxb-ri/issues/1287

jbonofre:
@ffang yes, I just would like to start the process, not include in 4.2.7 target.

PR #950, opened 2019-09-25
Updating Jackson to 2.9.10

coheigea:
We should update Jackson to the latest version due to CVE-2019-14540.


jbonofre:
Duplicated

PR #951, opened 2019-10-05
Added more information to the thrown exception for malformed url.

miroslav-beranic:
Adding more information to the stacktrace what is the problem with the URL and for what Feature it is the problem.

### DIFF
--- a/features/core/src/main/java/org/apache/karaf/features/internal/service/FeatureConfigInstaller.java
+++ b/features/core/src/main/java/org/apache/karaf/features/internal/service/FeatureConfigInstaller.java
@@ -103,7 +103,11 @@ public class FeatureConfigInstaller {
             // trim lines
             String val = config.getValue();
             if (config.isExternal()) {
-                props.load(new URL(val));
+                try {
+                    props.load(new URL(val));
+                } catch (java.net.MalformedURLException e) {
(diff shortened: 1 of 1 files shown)


miroslav-beranic:
https://issues.apache.org/jira/browse/KARAF-6447

miroslav-beranic:
Hi,

I was using this now for few days. I would suggest to to have it IOException, with addition of adding origin exception. Why?

Well, as is now ( change in PR ), "origin exception" is hidden, as I did not add the origin exception ( forgot about it ). So, the output does not contain origin error - that the "mvn:" prefix is not recognized.

With suggested solution, there is another problem. I agree exception MalformedURLException would be more suited, but it does not have constructor to accept origin exception, plus in this case it only re-throws exception.

Also, containing method declares as throws IOException, that is also in line with exception throws, so the user of this method can have less catch statements.

IOException is also correct, because of the higher-level exception, that is a result of MalformedURLException -- think of it as: due to malformed URL, file was not found and could not be read ( IOException ).

Based on all this, I suggest and think, that the exception IOException is the correct one to throw, but it is missing additional argument - origin exception, that is MalformedURLException and visible in a stacktrace, also accessible if needed.

Please let me know what you think about this.

Kind Regards,
Miroslav

ffang:
@miroslav-beranic , your explainition sounds good. Could you please also add  origin MalformedURLException when construct the IOException you want to throw?
Thanks!

PR #952, opened 2019-10-08
[KARAF-5628] Corrupt gc.log due to unseparated VM settings

avasilev:
setenv would recognize original calling script. So EXTRA_JAVA_OPTS or JAVA_OPTS could be set from there script based with bat/sh if statement.


jbonofre:
retest this please

PR #955, opened 2019-10-15
Put JAVA_OPTS between quotes for handling parameters with spaces

mdonkers:
Certain Java options might contain spaces, for example the `-XX:OnError` which can be used to execute a command or script and takes the parameter `%p` as placeholder for the current PID of the Java process. However, due to the way e.g. Bash handles variables in scripts, it makes a lot of difference whether these are placed between quotes. Without quotes, spaces are handled so that they break up parameters.
This causes e.g. the following `export JAVA_OPTS="-XX:OnError=\"echo %p\""` to be handled as two separate arguments resulting in the error: `Error: Could not find or load main class %p`

The only downside I found from quoting is if e.g. a user has a script `export JAVA_OPTS="${JAVA_OPTS} ${MY_OPTS}"` and `JAVA_OPTS` was not previously set, an additional space is introduced which is not handled nicely.

### DIFF
--- a/assemblies/features/base/src/main/filtered-resources/resources/bin/karaf
+++ b/assemblies/features/base/src/main/filtered-resources/resources/bin/karaf
@@ -299,7 +299,7 @@ run() {
 
         if [ "${ROOT_INSTANCE_RUNNING}" = "false" ] || [ "${CHECK_ROOT_INSTANCE_RUNNING}" = "false" ] ; then
             if [ "${VERSION}" -gt "8" ]; then
-                ${KARAF_EXEC} "${JAVA}" ${JAVA_OPTS} \
+                ${KARAF_EXEC} "${JAVA}" "${JAVA_OPTS}" \
                     --add-reads=java.xml=java.logging \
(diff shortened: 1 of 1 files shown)


jbonofre:
This patch is actually bad as it breaks the startup:

```
./karaf 
Invalid initial heap size: -Xms128M -Xmx512M -XX:+UnlockDiagnosticVMOptions  -Dcom.sun.management.jmxremote
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

I'm fixing.

jbonofre:
Why not simply breaking space with \\ ? It should work with the original script.

mdonkers:
I've tried adding `\\` for escaping spaces. Unfortunately that doesn't work in scripts because it gets parsed and when passed along the `\\` escaping is removed.

I had mentioned the possible breaking:
> The only downside I found from quoting is if e.g. a user has a script export JAVA_OPTS="${JAVA_OPTS} ${MY_OPTS}" and JAVA_OPTS was not previously set, an additional space is introduced which is not handled nicely.

So somewhere it must introduce additional spaces which then the JVM strangely can't handle.
I'll check again if there is a way around that. Seems like whichever way you go, things break.

PR #958, opened 2019-10-16
[KARAF-6458] Add new ignorePartialResultException property to jaas-ldap

coyotesqrl:
This property, when set to `true` and paired with a `false` value for `context.java.naming.referral` (the default for that property), will ignore `PartialResultException`s thrown from Active Directory when referrals are nested beneath role base dns.

This follows the same solution model as spring-ldap, as discussed in the [ticket here](https://issues.apache.org/jira/browse/KARAF-6458#).

Assuming this solution is accepted, I would also like it backported to the 4.2.x branch.


jbonofre:
retest this please

PR #978, opened 2019-10-30
[KARAF-6452] Add SCR CXF example

jbonofre:


jbonofre:
Test failure is unrelated.

PR #993, opened 2019-11-11
interceptor support on top on services registry

rmannibucau:
Goal of this PR is to provide an interceptor support on top of SCR.
The idea is provided in tests (I didnt write yet an e2e test, only the classes to test, this is why this PR is "WIP")  ->  see karaf/service/interceptor/impl/src/test/java/org/apache/karaf/service/interceptor/impl/test. Foo is a service marked with @Interceptors which is a holder for a list of interceptor marker (to not bind the interceptor impl in the consumer directly and be able to change it if needed, it will likely be enriched with other metadata but this is a first version). Then WrappingInterceptor defined the actual interceptor with the same marker.

The API is inspired from JavaEE but is simplified and limited to the minimum. This should enable to implement easily bridges (I'm thinking to microprofile typically) without having to require javax.interceptor and friends (in particular cause it will all change to jakarta.interceptor very soon).

Hope it makes sense.
Will try to see how to write an E2E soon but wanted to share it early (a mail should follow up on the list very shortly).

### DIFF
--- a/manual/src/main/asciidoc/user-guide/interceptor.adoc
+++ b/manual/src/main/asciidoc/user-guide/interceptor.adoc
@@ -1,0 +1,112 @@
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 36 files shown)


jbonofre:
@rmannibucau can you please squash the commit ? Is it ready for full review (not a WIP) ? Thanks !

rmannibucau:
@jbonofre yep, ready for review now (long story short, it can now be enhanced - I am thinking to binding parameters for example - without breaking the API so for a first delivery it is good enough IMHO).
Ill squash as soon as I get a computer today.

jbonofre:
retest this please

jbonofre:
LGTM, ready to merge if you agree.

rmannibucau:
+1, we can still move it later

PR #995, opened 2019-11-18
[KARAF-5770] OSGi R7 Support

jbonofre:

PR #1000, opened 2019-11-26
startup bundle level support

rmannibucau:
For master and R7 branch, goal is to ensure maven plugin startup bundle order is respected (since start level is not yet mapped in the config)


jbonofre:
last commit cherry-pick on `master` and `karaf-4.2.x` and rebase on #995

PR #1001, opened 2019-11-27
Update pax url

coheigea:


jbonofre:
It's already convered by another PR. Thanks.

PR #1004, opened 2019-12-12
add feature whitelist definition

mhus:
Add the node whitelistedRepositories in org.apache.karaf.features.xml and define features as accepted before the balcklist filter will drop the feature


mhus:
Found a small bug!

PR #1020, opened 2019-12-22
add feature whitelist definition

mhus:
Add the node whitelistedRepositories in org.apache.karaf.features.xml and define features as accepted before the blacklist filter will drop the feature, also add feature to karaf maven plugin


jbonofre:
I think I saw another PR related to same thing. Does this one supersede the previous one ?

mhus:
I added support for whitelists in the maven plugin too and had to reorganise the fork. Learned not to use a master for pull requests. This is a replacement for pull request 1005.

PR #1039, opened 2020-01-02
[KARAF-6574] Update pax-logging dependency to 1.11.4 and log4j2 to 2.13.0 (karaf-4.2.x)

mattrpav:
1.11.4-SNAPSHOT for now to run tests. Will ssquash PR when pax-logging 1.11.4 is released


jbonofre:
No need to create PR for other branches than `master`.

PR #1068, opened 2020-02-13
Add osgi.core in the instance classpath (required by pax-logging)

coshxx:
Basically same issue as https://github.com/apache/karaf/pull/1045


jbonofre:
Good catch, I fixed in `bin/client` but forgot instance. Thanks.

PR #1079, opened 2020-03-21
[KARAF-6650] Fix NullpointerException parsing bundle.info file

oklimberg:


oklimberg:
Yeah, sure. 
I can just force push the squashed changes right?
FYI Your contribution guide states, squashing the changes for a pull request is not mandatory. Should this be changed?

jbonofre:
It’s not mandatory when each commit makes sense. In your case the three commits are related to the same thing/change.

oklimberg:
Ah, okay, got it. Thx.

PR #1081, opened 2020-03-24
fix filter to filter over group and artifact

mhus:
Small fixes in service plugin...
* fix filter to filter over group and artifact (groupId:artifactId) - backward compatible
* fix output to print service class - for better analysis


jbonofre:
Can you please elaborate a bit the issue you are fixing here (with a quick test case to understand, it would be helpful) ?

mhus:
The implementation was filtering only over the artifact id. The fix enhances the filter to check [groupid]:[artifactid] against the regex include filter.

I also enhanced the output to show more details.

PR #1083, opened 2020-04-03
KARAF-6661: Add javax.activation to jre9

graben:


ffang:
Please don't merge, we have some reasons to do it
https://issues.apache.org/jira/browse/KARAF-4154

graben:
@ffang : Does this mean the javax.activation-1.2.0.jar under lib/jdk9plus is wrong?

ffang:
Hi @graben ,

No, it's not wrong, IIRC it's on the classpath so it's still be used to launch the OSGi container at early stage, when OSGi classloading(bundle export) mechanism not kick in yet.

You should install activation bundle yourself to use it.

Best Regards
Freeman

graben:
So it's needed to be bundled twice.

jbonofre:
Related to that, you can see the message I sent on the mailing list couple of weeks ago: `jdk9plus` should be removed and we should preferably use spec bundles.

PR #1096, opened 2020-04-11
KARAF-6676 Support OSGi R7 factory configurations with factory PID an…

oliverlietz:
…d name


jbonofre:
Good catch, I forgot to update in the features service. Thanks !

jbonofre:
I'm doing the review.

oliverlietz:
@jbonofre, Please review also the use of `CONFIG_KEY` (`org.apache.karaf.features.configKey`) in `findExistingConfiguration`. Doesn't make sense to me (yet).

PR #1133, opened 2020-05-26
[KARAF-6731] Align hibernate bundle version with feature version

jbonofre:


jbonofre:
JPA example has to be updated (missing requirement).

PR #1141, opened 2020-08-10
Fix unnecessary bundle refreshes when installing/updating features

badfish69:
Currently, the feature deployer uses the OSGI Resource interface object to determine dependent bundles that require refreshing when bundles associated with a feature are installed or updated.  Unfortunately, there are a number of conditions where equivalent resources will not satisfy the equality conditions of the Resource implementation class(es) involved, resulting in unnecessary dependent bundle refreshes during deployment.

This change replaces the use of the Resource interface with a wrapper class to ensure that equality is evaluated solely on symbolicName/version.

### DIFF
--- a/features/core/src/main/java/org/apache/karaf/features/internal/service/Deployer.java
+++ b/features/core/src/main/java/org/apache/karaf/features/internal/service/Deployer.java
@@ -32,6 +32,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


skitt:
@rovarga this might interest you...

jbonofre:
Just a side note, it looks good to me for the "current" feature resolver. The coming simple feature resolver (optional as an alternative to the regular one) is doing a "flat" resolution (no auto refresh).

jbonofre (approved):

PR #1142, opened 2020-08-10
[KARAF-6797] More detailed description of the JAAS command options fo…

AndreVirtimo:
The option "--realm" of the command jaas:realm-manage have to be combine with the option "--module" and vice versa. This wasn't mentioned in the help message of the command.

### DIFF
--- a/jaas/command/src/main/java/org/apache/karaf/jaas/command/ManageRealmCommand.java
+++ b/jaas/command/src/main/java/org/apache/karaf/jaas/command/ManageRealmCommand.java
@@ -35,14 +35,14 @@ import org.apache.karaf.shell.api.action.lifecycle.Service;
 @Service
 public class ManageRealmCommand extends JaasCommandSupport {
 
-    @Option(name = "--realm", description = "JAAS Realm", required = false, multiValued = false)
+    @Option(name = "--realm", description = "JAAS Realm. Must be combined with --module", required = false, multiValued = false)
     @Completion(RealmCompleter.class)
     String realmName;
 
     @Option(name = "--index", description = "Realm Index", required = false, multiValued = false)
     int index;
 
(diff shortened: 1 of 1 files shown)


jaas/command/src/main/java/org/apache/karaf/jaas/command/ManageRealmCommand.java
```diff
@@ -35,14 +35,14 @@
 @Service
 public class ManageRealmCommand extends JaasCommandSupport {
 
-    @Option(name = "--realm", description = "JAAS Realm", required = false, multiValued = false)
+    @Option(name = "--realm", description = "JAAS Realm. Must be combined with --module", required = false, multiValued = false)
```
jbonofre, line 38: That's not fully correct. Realm can be combined with module, but also with index.

jaas/command/src/main/java/org/apache/karaf/jaas/command/ManageRealmCommand.java
```diff
@@ -35,14 +35,14 @@
...
     @Completion(RealmCompleter.class)
     String realmName;
 
     @Option(name = "--index", description = "Realm Index", required = false, multiValued = false)
```
jbonofre, line 42: I would add the "must be combined" statement here and leave realm option description as it is.

jaas/command/src/main/java/org/apache/karaf/jaas/command/ManageRealmCommand.java
AndreVirtimo: Why should I combine this with index? Index is already a unique attribute.

jbonofre: Ah yes, my bad, I changed that. You are right.

PR #1066, opened 2020-02-08
Add karaf-rest-example with Jakarta Jersey client

benroe89:

PR #1153, opened 2020-08-16
[Examples] remove "cxf" context path from REST-endpoint URLs

benroe89:
[Examples] remove "cxf" context path from REST-endpoint URLs

...as publishing done to / for now on.


jbonofre:
Please use #1066 for this change. Anyway, it should work with CXF and cxf will be part of the URL when using CXF.

PR #1154, opened 2020-08-17
[KARAF-6815] Upgrade to maven-bundle-plugin 5.1.1

jbonofre:


jbonofre:
I'm fixing the `KarMojo` to use the "new" methods signature.

PR #1155, opened 2020-08-17
[KARAF-6624] Add JDK13 & JDK14 support

jbonofre:


jbonofre:
I have to check `etc/jre.properties` for JDK 13/14.

PR #1162, opened 2020-08-20
Update README.md

PoojaChandak:
Observed a few changes. Suggesting the same. Kindly check


oscerd (approved):

skitt:
Thanks for your contribution! Could you explain what this PR fixes? The text you’re changing is correct English, as is your new text. What are the changes that you observed?

PoojaChandak:
Hi, This PR is suggesting an alternative way of expressing the context. I understand the existing isn't wrong, however, I thought the suggestions might be an acceptable request.

jbonofre (changes requested):
Sorry, but I don't see the value of this PR. If the changes are part of a larger proposal, I can understand, but just "wording" change is creating noise for nothing to me.

PR #1164, opened 2020-08-20
whitespace

gmcdonald:
test PR, so not merge

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
 [Apache Karaf](https://karaf.apache.org) is a modern and polymorphic applications runtime.
 It's a lightweight, powered, and enterprise ready runtime powered by OSGi.
 [Apache Karaf](https://karaf.apache.org) is a "product project", providing a complete and turnkey runtime. The runtime is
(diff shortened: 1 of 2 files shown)


skitt:
@gmcdonald you can convert this to a draft, that way it won’t risk being merged ;-). All the checks should still run. On the right-hand-side, in the “Reviewers” section, click on “Still in progress? Convert to draft”.

gmcdonald:
> @gmcdonald you can convert this to a draft, that way it won’t risk being merged ;-). All the checks should still run. On the right-hand-side, in the “Reviewers” section, click on “Still in progress? Convert to draft”.

Now why didnt I think of that!

gmcdonald:
I am wondering if the problem is that this is a Maven type job and cant be used with GH App creds.

Any way you can copy this job into a multibranch or Org Folder type job?

jbonofre:
@gmcdonald we already have a multibranch project on new-builds as well. However, I wanted to avoid to create multibranch/org folder to mimic what we had before on buids.a.o.

PR #1167, opened 2020-08-23
Add jenkins file into a new branch

gmcdonald:


jbonofre:
Applied on `master` (and so it will be included on future PRs), it works like a charm ! Thanks !

PR #1168, opened 2020-08-26
Export JAVA_OPTS after setting EXTRA_JAVA_OPTS

htreu:
Export JAVA_OPTS after adding EXTRA_JAVA_OPTS.


oscerd (approved):

skitt (approved):
This doesn’t make any technical difference, but it does make the intent clearer.

(`export` sets a flag indicating that the given variable should be exported; it doesn’t export it there and then. When a new process is started, it gets an environment constructed including all exported variables, with their values at that moment, not the value when they were marked as exported.)

htreu:
Thanks for the explanation @skitt, nice to know.

jbonofre (approved):
I don't see any difference ;) I will merge, thanks.

PR #1169, opened 2020-08-28
Enhance RunMojo 

rmannibucau:
to let it configure main args, console log level, system properties and be able to deploy a bundle before/after the features with local lookup instead of assuming the artifact is attached to the project (karaf:run without any other goal)


fpapon:
@rmannibucau thanks for the PR!

PR #1181, opened 2020-09-04
Fix link to project page

AndreVirtimo:
Found some dead links. Currently there is no page for each project, so I linked to the projects page, where all subprojects are listed.


jbonofre (approved):
Thanks ! Good catch !

PR #1191, opened 2020-09-16
KARAF-6852 - RmiServerPort (44444) does not rise at startup

coheigea:


oscerd (approved):

jbonofre (approved):
I tested successfully. I will add a itest checking if the port are bound, just to avoid such regression in the future.
Thanks !

PR #1201, opened 2020-09-25
add support of clean arg in main

rmannibucau:


main/src/main/java/org/apache/karaf/main/Main.java
```diff
@@ -232,6 +239,29 @@ public void updateInstancePidAfterShutdown() throws Exception {
     }
 
     public void launch() throws Exception {
+        if (Arrays.asList(args).contains("clean")) {
```
jbonofre, line 242: It does already exist. You can start `bin/karaf` with `clean` option. That's in the `sh|bat` script.
So your proposal is to remove from script to have it in `Main`, right ?
In that case, we should also upgrade the scripts.

jbonofre (changes requested):

main/src/main/java/org/apache/karaf/main/Main.java
rmannibucau: Exact need is to support this feature when no using the scripts (maven plugin, gradle, embedded case). +1 to clean the scripts - just need to check it does not fail more often on windows (saw java deletion code not that reliable some years ago)

jbonofre:
Superseeded by #1214

PR #1215, opened 2020-10-06
Alternatively read features.core configuration from ConfigAdmin

htreu:
* work around https://issues.apache.org/jira/browse/KARAF-6866

In rare cases we see features.core Activator to read an empty `etc/org.apache.karaf.features.cfg`, probably a race with felix FileInstall. This works around this issue by reading the properties alternatively from ConfigurationAdmin which gets a copy upfront.


htreu:
@jbonofre I added a test case basically running the new code. Unfortunately sonar will not recognise it, any hints appreciated.

jbonofre:
@htreu don't worry about Sonar for now ;)

PR #1221, opened 2020-10-13
circular buffer extraction was not accurate

rmannibucau:


jbonofre:
Tested succesfully thanks !

PR #1226, opened 2020-10-15
[KARAF-6887] Remove default JAVA_MAX_MEM JAVA_MIN_MEM 

mattrpav:
settings from bin/inc


jbonofre (approved):

jbonofre:
For the record:
1. please create PR based on `master` if it's common and can be cherry picked to `karaf-4.2.x`
2. please create PR  only containing commits specific to the problem you address and squash all in one commit

PR #1244, opened 2020-10-28
Update pax-logging-version

tobias-schanze:
Update pax-logging-version to 1.10.7, because in 1.10.1 stacktraceAsString doesn't work.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -259,7 +259,7 @@
 
         <pax.cdi.version>1.0.0.RC2</pax.cdi.version>
         <pax.exam.version>4.11.0</pax.exam.version>
-        <pax.logging.version>1.10.1</pax.logging.version>
+        <pax.logging.version>1.10.7</pax.logging.version>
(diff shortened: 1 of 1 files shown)


oscerd (changes requested):
I do not think it needs to be bump on old karaf version.

tobias-schanze:
> I do not think it needs to be bump on old karaf version.

I think it would nice.. Karaf 4.1.x is almost a stable version..

oscerd:
> > I do not think it needs to be bump on old karaf version.
> 
> I think it would nice.. Karaf 4.1.x is almost a stable version..

I believe the policy is to bump only for CVE or particular situation. Btw let's wait for other opinions

jbonofre:
We encourage users to update to 4.2.x or 4.3.x (as it's now on vote).
I will merge this one, and eventually cut a release, but again, please upgrade ;)

jbonofre (approved):

PR #1256, opened 2020-11-24
Fix async logging comment message

harrison-tarr:
I was trying to enable asynchronous logging and found that with just the com.lmax/disruptor dependency I was getting a "NoClassDefFound" exception. With some research, I found I needed to add the `pax-logging-log4j2-extra` artifact. Now my custom distro starts


jbonofre:
@harrison-tarr thanks for the update.

oscerd (approved):

jbonofre (approved):

PR #1259, opened 2020-11-27
KARAF-6925 - Initial attempt at supporting spring-security-crypto

coheigea:
Some points to note:
 * In jaas/spring-security-crypto/pom.xml it just depends on a servicemix bundle, I wasn't sure whether to just depend on the normal spring jar here.
 * The feature I added could be reworked to remove dependency on a set version of Spring.
 * scrypt, bcrypt, pbkdf2, argon2 work well.
 * Remaining work: Update the default jaas file comments to show how to use it, and update the documentation.

### DIFF
--- a/assemblies/features/standard/pom.xml
+++ b/assemblies/features/standard/pom.xml
@@ -390,6 +390,13 @@
             <scope>provided</scope>
         </dependency>
 
+        <!-- Spring Security Crypto encryption deps -->
+        <dependency>
+            <groupId>org.apache.karaf.jaas</groupId>
+            <artifactId>org.apache.karaf.jaas.spring-security-crypto</artifactId>
+            <scope>provided</scope>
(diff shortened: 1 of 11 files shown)


jbonofre:
I would use a different approach embedding spring-security as private to avoid feature dependency. Let me rework this PR.

jbonofre:
As said, I will rework this one to avoid Spring bundles dependency and prefer spring-security embedded. I will do that tomorrow. Thanks !

coheigea:
Thanks @jbonofre, I was just making a trivial change to the logic.

jbonofre:
As it's an optional JAAS bundle, it makes sense to keep it as a feature (no need to use private package as it's not part of "core" JAAS). LGTM.

jbonofre (approved):

PR #1261, opened 2020-12-07
Add ability to set additional non-debug JVM options

fwolter:
Introduces a new environment variable to set additional JVM options only when not running in debug mode.

The current implementation only provides variables for both, running in debug mode and non-debug mode and for running in debug mode, but not for running in non-debug mode only.

The use case at openHAB is to disable the JVM performance logging when running normally as the periodic disk writing kills flash memory e.g. in raspberry PIs. But it may not be disabled when running in debug mode as it kills the ability to use VisualVM. See openhab/openhab-distro#1183 and openhab/openhab-distro#1193.

### DIFF
--- a/assemblies/features/base/src/main/filtered-resources/resources/bin/karaf
+++ b/assemblies/features/base/src/main/filtered-resources/resources/bin/karaf
@@ -273,6 +273,8 @@ run() {
             fi
         fi
         JAVA_OPTS="${JAVA_DEBUG_OPTS} ${JAVA_OPTS}"
+    else
+        JAVA_OPTS="${JAVA_NON_DEBUG_OPTS} ${JAVA_OPTS}"
     fi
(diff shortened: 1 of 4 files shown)


jbonofre:
Thanks for the contribution. Just curious: why can't we use `JAVA_OPTS` directly ?

fwolter:
Thanks! `JAVA_OPTS` is valid for both debug and non-debug. We are disabling logging of performance data only when running in non-debug.

jbonofre:
Understood, it's really specific only when not using debug. It makes sense, thanks !

jbonofre (approved):

jbonofre:
I also cherry picked on `karaf-4.2.x` branch. Thanks again !

fwolter:
Thanks!

PR #1267, opened 2020-12-15
[KARAF-6917] Remove -Dcom.sun.management.jmxremote by default

jbonofre:


oscerd (approved):

manual/src/main/asciidoc/user-guide/monitoring.adoc
```diff
@@ -32,14 +32,6 @@ The JMX URL to use by default is:
...
 ----
 
-If you don't need the remote JMX at all, users can remove
-
-----
--Dcom.sun.management.jmxremote
-----
-
-from bin/karaf|bin/karaf.bat to avoid opening the RMI listening port.
-
```
coheigea, line 42: Probably here we need to add documentation to put in -Dcom.sun.management.jmxremote if remote monitoring is needed?

jbonofre: We can do that anyway, can't we ? jconsole, decanter JMX, etc, can still connect for remote monitoring.

PR #1269, opened 2020-12-15
[KARAF-6963] First send SIGTERM, only SIGKILL if not shutting down nicely

mdonkers:
PR for issue: https://issues.apache.org/jira/browse/KARAF-6963

Try first a regular `SIGTERM` to shutdown the daemon process.
If that doesn't result in the process stopping within 10 seconds, follow with the `SIGKILL` signal.

It's verifying in 1 second intervals whether the process is still running, so that the 'stop' request won't be blocking too long.

### DIFF
--- a/instance/src/test/java/org/apache/karaf/jpm/MainTest.java
+++ b/instance/src/test/java/org/apache/karaf/jpm/MainTest.java
@@ -19,6 +19,18 @@ package org.apache.karaf.jpm;
 public class MainTest {
 
     public static void main(String[] args) throws Exception {
+        // Added shutdown hook to block after SIGTERM
+        Runtime.getRuntime().addShutdownHook(new Thread() {
+            @Override
+            public void run() {
+                try {
+                    Thread.sleep(Long.parseLong(args[1]));
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
+                }
+            }
+        });
+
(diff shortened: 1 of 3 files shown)


mdonkers:
Rebundling our code with Karaf and with this fix applied, the exit status is now 'clean':

```
root@vagrant:/opt/instana# systemctl stop instana-agent
root@vagrant:/opt/instana# systemctl status instana-agent
● instana-agent.service - "Instana(tm) agent."
   Loaded: loaded (/lib/systemd/system/instana-agent.service; enabled; vendor preset: enabled)
   Active: inactive (dead) since Wed 2020-12-16 10:27:35 UTC; 3s ago
  Process: 14272 ExecStop=/opt/instana/agent/bin/karaf stop (code=exited, status=0/SUCCESS)
  Process: 14180 ExecStart=/opt/instana/agent/bin/karaf daemon (code=exited, status=0/SUCCESS)
 Main PID: 14180 (code=exited, status=0/SUCCESS)
```

mdonkers:
*Bump*.
I haven't seen any updates. I think the build error is not correct or at least not related to this change, but cannot re-trigger.
Would be nice if I could get some feedback, to move this forward.
Many thanks.

jbonofre:
Thanks for the PR. I'm reviewing it now.

jbonofre (approved):

PR #1281, opened 2020-12-24
[KARAF-6703] Add specs features repo

jbonofre:


jbonofre:
This is a first round, base to be improved in next releases and used by third parties.

PR #1287, opened 2021-01-08
[KARAF-6953] Introduce autoRefresh property in etc/org.apache.karaf.features.cfg

jbonofre:


features/core/src/main/java/org/apache/karaf/features/FeaturesService.java
```diff
@@ -42,6 +42,7 @@
     long DEFAULT_SCHEDULE_DELAY = 250;
     int DEFAULT_SCHEDULE_MAX_RUN = 9;
     long DEFAULT_REPOSITORY_EXPIRATION = 60000; // 1 minute
+    boolean DEFAULT_AUTO_REFRESH = false;
```
daniellas, line 45: Does it change current default setting?

k-brooks:
Thanks!

jbonofre:
I handle the `deploy` phase, I will do the same for the `undeploy` phase.

ahgittin:
Nice one!  Looking forward to this.

PR #1318, opened 2021-03-05
[KARAF-7061] Change log4j2.pattern to prevent CRLF/HTML code injection

jbonofre:


sergehuber (approved):
Thank you so much for doing this so quickly, really appreciated !

PR #1346, opened 2021-04-11
KARAF-7099 Provide util for Configuration PIDs

oliverlietz:


jbonofre:
It would be great to have a quick abstract / description of the PR. Thanks.

util/pom.xml
```diff
@@ -74,6 +74,13 @@
             <artifactId>mail</artifactId>
             <scope>test</scope>
         </dependency>
+        <!-- nullability -->
+        <dependency>
+            <groupId>org.jetbrains</groupId>
+            <artifactId>annotations</artifactId>
```
jbonofre, line 80: Do we really need this dependency in util for null check ?

jbonofre (changes requested):

util/pom.xml
oliverlietz: In general I prefer to annotate parameters and return values but the dependency is of course not strictly required. The annotations (https://github.com/JetBrains/java-annotations) are used during analysis and compile time only and should have no negative impact.
Should I remove the annotations from ConfigurationPID?

jbonofre: Don't worry, I'm reworking your two PRs. Thanks !

jbonofre:
Superseeded by #1368 1368

PR #1375, opened 2021-05-08
[KARAF-7097] Fix json config array comparison

jbonofre:


oscerd (approved):

oliverlietz:
`updated` must never become `false` again once `true`:

```
            for (String key : old.keySet()) {
                Object oldValue = old.get(key);
                Object propertiesValue = properties.get(key);
                if (oldValue instanceof Object[] && propertiesValue instanceof Object[]) {
                    if (!Arrays.deepEquals((Object[]) oldValue, (Object[]) propertiesValue)) {
                        updated = true;
                    }
                } else {
                    if (!oldValue.equals(propertiesValue)) {
                        updated = true;
                    }
                }
            }
```

PR #1377, opened 2021-05-15
KARAF-7152 Fixing Narayana object store initialisation

PiotrKlimczak:


oscerd (approved):

PiotrKlimczak:
I am using MacOS and I have a feeling that due to missing AIO lib in MacOS, Narayana might display slightly different logs in Linux and other systems. Don't have any Linux around unfortunately.
It is test passing on my machine every time.

PiotrKlimczak:
Pushed assertion fix which I hope should not be AIO dependant anymore.

PiotrKlimczak:
OK, I know what the problem is, unit test is hitting https://issues.apache.org/jira/browse/FELIX-6416 which I have reported separately, have it fixed locally but it is not yet merged.
I have raised separately https://issues.apache.org/jira/browse/KARAF-7153 as blocker for this issue.
Test will keep failing until https://issues.apache.org/jira/browse/KARAF-7153 is resolved.

PR #1379, opened 2021-05-23
KARAF-7155 - Deleting a factory configuration doesn't delete the associated json file

jassuncao:
JsonConfigInstaller keeps tracks of configurations that are associated
to a json file. When the configuration is deleted it will try to delete
the file.


jbonofre:
@jassuncao what about using `felix.fileinstall.filename` to delete the file ? This way we don't have to store the factory/pid in the map, just read the `felix.fileinstall.filename` from the properties and remove it.
It's the fix I did.
Thoughts ?

jassuncao:
@jbonofre I had the idea configuration properties wouldn't be available after receiving a ConfigurationEvent.CM_DELETED. I'm almost sure I attempted that, but I might have missed something,

jbonofre:
@jassuncao thanks, let me try something and I will let you know.

jbonofre:
Actually, it works for me.

PR #1385, opened 2021-05-31
[KARAF-7167] Upgrade to commons-io 2.10.0

jbonofre:


oscerd (approved):

jbonofre:
commons-io needs a new import which actually should be optional (`com.sun.nio`).

jbonofre:
commons-io 2.10.0 will be released soon with the fix on the OSGi headers.

PR #1394, opened 2021-06-23
[KARAF-7186] Optionally avoid to fail the verify goal on bundle uninstall/update

mrulli:
It would be useful for us to cherry pick this on Karaf 4.2.x line as well. Thanks


jbonofre (approved):
LGTM, thanks, and agree to backport on Karaf 4.2.x as it doesn't impact the existing.

PR #1419, opened 2021-08-17
[KARAF-7240] Upgrade to bouncycastle 1.68

jbonofre:


oscerd (approved):

pom.xml
```diff
@@ -169,7 +169,7 @@
         <asm.version>9.2</asm.version>
         <javax.annotation.version>1.3.1</javax.annotation.version>
         <awaitility.version>3.1.6</awaitility.version>
-        <bouncycastle.version>1.66</bouncycastle.version>
+        <bouncycastle.version>1.68</bouncycastle.version>
```
skitt, line 172: Can’t we upgrade to 1.69? That’s the current release.

```suggestion
        <bouncycastle.version>1.69</bouncycastle.version>
```

skitt (changes requested):

pom.xml
jbonofre: @skitt no, we can't as the OSGi header in 1.69 are buggy. I have to fix that at bouncycastle.

skitt: Ah, right, so they are...

skitt (approved):

PR #1449, opened 2021-11-07
Speedup calculating distances on Feature Resource digraph

globalbus:
I always blamed felix resolver for slow dependency resolution, but code stucking place is within karaf itself.
Current implementation sorts unSettledNode list on each loop iteration, so the computional complexity is horrible. For big features with hunderds of bundles, that makes a big difference.

It's only an implementation change, could be safety backmerged for older versions.


jbonofre:
@globalbus thanks for this proposal, it's a great one. Let me quickly review/test and I will merge. Thanks again !

grgrzybek:
wow - thanks!

jbonofre (approved):

PR #1456, opened 2021-11-26
KARAF-7282: fix autoRefresh setting treatment in Deployer

swachter:
Issue: https://issues.apache.org/jira/browse/KARAF-7282


oscerd (approved):

jbonofre (approved):

jbonofre:
I reverted this change as the original code looks good to me.

swachter:
I do not understand why you reverted this. I tried the global setting `autoRefresh=false` and saw that auto refresh takes place nevertheless. Interestingly, [PR#1457](https://github.com/apache/karaf/pull/1457) seems to reintroduce the change.

jbonofre:
The original change breaks the build as the DeployerTest was not updated. In order to avoid to break all builds, I reverted and did a new fresh/clean PR with DeployerTest updated.

PR #1463, opened 2021-12-08
KARAF-7289 - Provide config property name completion using MetaType info

jassuncao:
ConfigurationPropertyCompleter now uses the MetaType service to retrieve
configuration property names in addition to the ones already defined in
the configuration.

Adds an utility method in MetaServiceCaller to find MetaType info for a
specified PID. This method accepts a function to do something with the
result. MetaCommand is refactored to use this function.


oscerd (approved):

jbonofre:
Thanks, it's a good idea. Let me review the changes.

jbonofre (approved):

PR #1474, opened 2022-01-03
[KARAF-7316] Support for standard maven coordinates.

splatch:
This commit introduces very basic support for semi standard group:artifact:type[:classifier]:version syntax.
In case if all parts are given they take over computed values.
If version is missing - computed one is used.
If type is missing - zip is assumed.
If classifier is missing (which is common) empty string will be used.

This should simplify handling of basic feature verification projects.

Signed-off-by: Łukasz Dywicki <luke@code-house.org>


splatch:
@jbonofre Please review, this makes verify mojo working with own defaults, without listing karaf framework/distro in project dependencies.
Current logic is broken as described in refereed issue [KARAF-7316](https://issues.apache.org/jira/browse/KARAF-7316).

jbonofre (approved):

PR #1477, opened 2022-01-05
[KARAF-7317] Support spaces in boot features path

jdmcmahan:
`URI.create(String)` throws a `URISyntaxException` if any of the resolved URIs from `featuresRepositories` contains a space.

This is common if customizing the boot features via the `karaf-maven-plugin`, then running the assembly from within a path containing spaces.

For example:

```xml
<plugin>
    <groupId>org.apache.karaf.tooling</groupId>
    <artifactId>karaf-maven-plugin</artifactId>
    <version>${karaf.version}</version>
    <configuration>
        <bootFeatures>
            <feature>standard</feature>
            <feature>my-feature</feature>
        </bootFeatures>
        ...
    </configuration>
</plugin>
```

will result in a new feature in `etc` as well as an entry in `etc/org.apache.karaf.features.cfg`:

```
featuresRepositories = file:${karaf.etc}/2b260482-45bc-427c-af65-8909c1f2f7a3.xml
```

However, if `karaf.etc` contains a space at runtime, Karaf fails to start because the URI cannot be parsed.

This PR simply replaces all spaces with `%20` before passing the path to `URI.create(String)`. The `encodePath` method can be further enhanced if other characters prove to be problematic.

### DIFF
--- a/features/core/src/main/java/org/apache/karaf/features/internal/service/BootFeaturesInstaller.java
+++ b/features/core/src/main/java/org/apache/karaf/features/internal/service/BootFeaturesInstaller.java
@@ -16,21 +16,15 @@
  */
 package org.apache.karaf.features.internal.service;
 
-import java.io.File;
-import java.net.URI;
-import java.util.ArrayList;
-import java.util.EnumSet;
-import java.util.Hashtable;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Set;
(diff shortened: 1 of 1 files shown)


jdmcmahan:
I have created a companion ticket at [KARAF-7317](https://issues.apache.org/jira/browse/KARAF-7317).

jbonofre (approved):

PR #1484, opened 2022-01-09
[KARAF-7325] Support custom properties to simplify adjustment of assemblies

splatch:
I've made a small adjustments in regular expressions, in all places I found related to it, so they work properly with additional properties mentioned in KARAF-7325.

All integration tests pass without failure meaning that change should be harmless to standardized deployments while opening window of opportunity to everyone who needs a slight adjustment of boot packages, system packages or custom capabilities. This should also allow a simpler patching of jre properties for these who need such thing.


RangerRick:
Ooh, I realize this is languishing in PRs, but it would be really handy if it could be included.

splatch:
Dang, its quite aged PR, indeed it deserves a push.

jbonofre:
This PR is stale and on "old" 4.2.x branch. I'm closing, feel free to reopen one on `main` if needed.

PR #1489, opened 2022-02-09
[KARAF-7389] Prevent two threads (feature installer, CM Event Dispatc…

grgrzybek:
…her through fileinstall) writing the same config file.

This works for my case, however it's very hard to provide unit test for this.

1. CM update happens without `felix.fileinstall.filename`, so FileInstall won't try to read and write the file in CM_UPDATED listener when `org.osgi.service.cm.Configuration#update()` is called
2. `updateStorage()` is called with tmp file, highly minimizing the risk that FileInstall's `org.apache.felix.fileinstall.internal.DirectoryWatcher#install()` will be called with empty (or partial) file

### DIFF
--- a/features/core/src/main/java/org/apache/karaf/features/internal/service/FeatureConfigInstaller.java
+++ b/features/core/src/main/java/org/apache/karaf/features/internal/service/FeatureConfigInstaller.java
@@ -137,10 +137,10 @@ public class FeatureConfigInstaller {
                     cfg = createConfiguration(configAdmin, cid);
                     cfgProps.put(CONFIG_KEY, cid.pid);
                     properties.put(CONFIG_KEY, cid.pid);
+                    cfg.update(cfgProps);
                     if (storage != null && configCfgStore) {
                         cfgProps.put(FILEINSTALL_FILE_NAME, cfgFile.getAbsoluteFile().toURI().toString());
                     }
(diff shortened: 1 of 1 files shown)


oscerd (approved):

features/core/src/main/java/org/apache/karaf/features/internal/service/FeatureConfigInstaller.java
```diff
@@ -137,10 +137,10 @@ public void installFeatureConfigs(Feature feature) throws IOException, InvalidSy
                     cfg = createConfiguration(configAdmin, cid);
                     cfgProps.put(CONFIG_KEY, cid.pid);
                     properties.put(CONFIG_KEY, cid.pid);
+                    cfg.update(cfgProps);
```
jbonofre, line 140: Don't we want to keep the update there ?

grgrzybek: I moved the update up, so it's updated _without_ `felix.fileinstall.filename` property at this stage - this ensures that fileinstall (handling `CM_UPDATED`) won't read+write the file being written at the same time with `updateStorage()`

jbonofre (approved):

PR #1536, opened 2022-04-29
[KARAF-7423] Add logback configuration option

i-thiemann:
I've added an xml based configuration service implementation for logback configurations and corresponding unittests and documentation.


i-thiemann:
> Are we sure we want logback by default ? I would rather prefer to keep log4j2 style by default, and document how to have logback instead. Or eventually create another framework/static feature (with logback classifier for instance) to easily use logback.
> 
> Thoughts ?

The default option is still log4j2. The logback service just also exists now and can be included if needed.

i-thiemann:
I understand the confusion. I made a mistake in including pax-logging-logback in the wrong place in assembly I'll remove it and update my pull request

i-thiemann:
retest this please

jbonofre:
If you don't mind, I will squash the 5 commits in a single one.

i-thiemann:
> If you don't mind, I will squash the 5 commits in a single one.

That would be fine.

PR #1537, opened 2022-05-02
[KARAF-7431] Fix jre.properties

rovarga:
Add known packages up to and including JDK17


jbonofre (approved):

jbonofre:
It seems the Jira is not the right one. I guess it should be KARAF-7431 (instead of KARAF-7414). I will update the commit message.

jbonofre:
FYI, as mentioned in the Jira, this change breaks itests on `services/interceptor/impl`. I'm fixing there.

PR #1544, opened 2022-06-03
K8s static-dist.yaml has leftovers

ecki:
The arguments line seems to be from a test container using the “polinux/stress” container, they have no meaning for Karaf.

It would probably be possible to move “run” to the args array, but since I am not sure if it still uses /bin/sh -c in that case, I did not modify it.


jbonofre:
Good catch, my bad. Thanks for pointing this.

PR #1549, opened 2022-06-12
[KARAF-7444] Add missing java.lang.runtime in jre.properties

jbonofre:


assemblies/features/base/src/main/filtered-resources/resources/etc/jre.properties
```diff
@@ -46,6 +46,7 @@ jre-base= \
  java.lang.management, \
  java.lang.ref, \
  java.lang.reflect, \
+ java.lang.runtime, \
```
MrEasy, line 49: entry under jre-14 can/should then be removed, or?

MrEasy (changes requested):

MrEasy (approved):

PR #1551, opened 2022-06-17
Propose docker run commands with clean up tag in manual

essobedo:
## Motivation

When working with Docker it is important to ensure that once over everything is properly cleaned up otherwise it quickly gets space consuming

## Modification

* Add the clean up flag to the docker run commands in the manual

### DIFF
--- a/manual/src/main/asciidoc/user-guide/docker.adoc
+++ b/manual/src/main/asciidoc/user-guide/docker.adoc
@@ -68,14 +68,13 @@ docker build --build-arg KARAF_VERSION=4.2.0 -t "karaf:4.2.0" karaf
 * Run Karaf in interactive mode (using Karaf official docker image):
 
 ----
-docker run -i -t --name karaf apache/karaf:latest karaf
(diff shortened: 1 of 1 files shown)


jbonofre:
Not sure it makes such sense like this. Maybe just mentioning the option is enough.

essobedo:
Do you mean that you intentionally want to keep the container? For debugging purpose or to reuse it later? If you prefer to keep it maybe docker stop or Ctrl + C (as foreground mode is used here) are less risky than docker kill, don't you agree? And maybe a section to explain how to re-start the container could be added WDYT? Or if you rather prefer to keep it as it is, I understand, let's close it, it is only a small remark I had while reading the very nice and clear manual of the project, congrats for it BTW.
Let me know what you prefer?

jbonofre:
Yes it's intentional. For instance, a classic use case is to start a container, do some stuff on it and then stop it and tag to create an image. 

So imho it's better to keep the container by default. If the user really wants to remove it after run then he can do it on his own.

essobedo:
I see, so let's close it then, sorry for the noise

PR #1599, opened 2022-07-07
[KARAF-7389] Set felix.fileinstall.filename property in config installer

jbonofre:


jbonofre:
@grgrzybek as you did the initial change, can you please take a look ? I suspect this change cause issue in KARAF-7453 (and of course KARAF-7389)

grgrzybek:
Hmm. I still believe that updating a configuration **with** `FILEINSTALL_FILE_NAME` property will trigger a race condition - a moment later Karaf will continue writing the file (even if using tmpFile first) and FileInstall will get notified (because of `FILEINSTALL_FILE_NAME` property) and will attempt to write the file.

If (as described by J. Brébec in https://issues.apache.org/jira/browse/KARAF-6080) the case is that fileinstall sees existing file in etc and feature has `<config>`, this is probably something different. For KARAF-6080, `if (!cfgFile.exists())` should be enough - as in initial fix for KARAF-6080.

PR #1603, opened 2022-07-14
Karaf 2.4.x

guguri12:
How can I modify key exchange algorithms used by the Karaf SSH shell? I try to add "kexAlgorithms = diffie-hellman-group-exchange-sha256,diffie-hellman-group14-shal " in org.apache.karaf.shell.cfg file and restart karaf, but it can't change the value of default key exchange algorithms. Please kindly help to solve this issue.


jbonofre:
karaf-2.x is not maintained anymore.

grgrzybek:
Karaf 2.4 is not even handling `kexAlgorithms` property, which was added in 44323c27529c8668db6cc836f00d0bb75e2cb4d5 for Karaf 4.1.

PR #1604, opened 2022-07-22
KARAF-7522 Fix stack overflow error with duplicate features.

splatch:
Hey folks,
I nailed reason why we get stack overflows while making validation and it seems to be caused by subsystem eating its own tail. I made a fix which is proven to work, but now I need a bit of your feedback if proposed way is fine.

Few points - `Feature` equality uses name and version, ignoring actual feature contents, so we can rely on sets instead of lists making my fix redundant. Yet, it would require quite big update of code and could lead to some unexpected behavior changes.


jbonofre:
It would be great to open PR based on `main`. Then we cherry pick on active branches. Let me do the rebase to test it.

jbonofre (approved):

PR #1621, opened 2022-09-17
KARAF-7552: add tests for feature:repo commands

awrb:
Hi, it's my first contribution so for starters I added some tests to feature repo commands.
If the PR is fine then I'll gladly submit more later.

### DIFF
--- a/features/command/src/main/java/org/apache/karaf/features/command/RepoAddCommand.java
+++ b/features/command/src/main/java/org/apache/karaf/features/command/RepoAddCommand.java
@@ -32,13 +32,13 @@ public class RepoAddCommand extends FeaturesCommandSupport {
 
     @Argument(index = 0, name = "name/url", description = "Shortcut name of the features repository or the full URL", required = true, multiValued = false)
     @Completion(AvailableRepoNameCompleter.class)
-    private String nameOrUrl;
-    
+    String nameOrUrl;
+
     @Argument(index = 1, name = "version", description = "The version of the features repository if using features repository name as first argument. It should be empty if using the URL", required = false, multiValued = false)
-    private String version;
+    String version;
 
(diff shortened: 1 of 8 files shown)


jbonofre (commented):
Did you take a look on the itests ?

The feature service and commands are tested in the itest:

https://github.com/apache/karaf/blob/main/itests/test/src/test/java/org/apache/karaf/itests/FeatureTest.java

https://github.com/apache/karaf/tree/main/itests/test/src/test/java/org/apache/karaf/itests/features

Your tests are interesting but they don't really test the service as it's using mocks.

Thoughts ?

awrb:
@jbonofre  Ah I missed the itests somehow. They're certainly more useful than the tests with mocked FeaturesService. There are a few similar tests that I looked at before adding the ones in the PR (testing the command with mocks rather than pax-exam), e.g.
`org.apache.karaf.bundle.command.ListServicesTest`
`org.apache.karaf.features.command.ListFeaturesCommandTest`
`org.apache.karaf.config.command.EditCommandTest`
so based on these I thought this approach for testing commands is OK.

It seems to me that it's still useful to have tests like that, to check if the command is calling the proper service methods, passing the correct arguments etc. If you are working on the commands, they're also easier to run than pax-exam and can act as a preliminary test. But it's best for the maintainers to decide. 

Another question: is some code coverage tool used in the project? I only found jacoco in tooling/karaf-maven-plugin. Maybe there is sonarqube on jenkins or something? I'd like to use something to check where more test cases can be supplied.

jbonofre:
@awrb yes, I think the tests are interesting, just wanted to spot on the itests ;)

Yes, we are using jacoco for test coverage + sonar. You can see the sonar configuration in the Jenkinsfile: https://github.com/apache/karaf/blob/main/Jenkinsfile#L96

However, it's commented right now as almost all PR builds are failing due to that.

jbonofre (approved):

PR #1624, opened 2022-09-19
KARAF-7554 - add feature:status command

awrb:
It's a subtask from [KARAF-5193](https://issues.apache.org/jira/browse/KARAF-5193), which is pretty old but I think it's still relevant (at least feature:status was never implemented).

### DIFF
--- a/features/command/src/main/java/org/apache/karaf/features/command/completers/FeatureVersionCompleter.java
+++ b/features/command/src/main/java/org/apache/karaf/features/command/completers/FeatureVersionCompleter.java
@@ -1,0 +1,70 @@
+/**
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.karaf.features.command.completers;
+
+import org.apache.karaf.features.Feature;
+import org.apache.karaf.features.FeaturesService;
+import org.apache.karaf.features.Repository;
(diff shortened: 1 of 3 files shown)


features/command/src/main/java/org/apache/karaf/features/command/completers/FeatureVersionCompleter.java
```diff
@@ -0,0 +1,41 @@
...
+    public int complete(Session session, CommandLine commandLine, List<String> candidates) {
+        StringsCompleter delegate = new StringsCompleter();
+        String[] args = commandLine.getArguments();
+        // args look like this at this point: [feature:status, wrapper, '']
+        if (args.length >= 3) {
+            String featureArg = args[1];
+            try {
+                Feature[] features = featuresService.getFeatures(featureArg);
+                List<String> versions = Arrays.stream(features).map(Feature::getVersion).collect(Collectors.toList());
+                delegate.getStrings().addAll(versions);
```
awrb, line 32: I wanted the completer to work if there are multiple versions of a single feature available, but sadly `versions` is always a singleton here. In `FeatureReq` the latest version is returned:
```
    public Stream<Feature> getMatchingFeatures(Map<String, Map<String, Feature>> allFeatures) {
        Pattern pattern = Pattern.compile(name);
        Function<String, Optional<Feature>> func = featureName -> {
            Feature matchingFeature = null;
            if (pattern.matcher(featureName).matches()) {
                Map<String, Feature> versions = allFeatures.get(featureName);
                matchingFeature = getLatestFeature(versions, versionRange);
            }
            return Optional.ofNullable(matchingFeature);
        };
        return allFeatures.keySet().stream().map(func).filter(Optional::isPresent).map(Optional::get);
    }
```
It does not seem possible to get all versions via `FeaturesService`

oscerd (approved):

features/command/src/main/java/org/apache/karaf/features/command/completers/FeatureVersionCompleter.java
awrb:
```
 private List<String> getAllVersionsOfFeature(String feature, FeaturesService featuresService) throws Exception {
        List<String> versions = new ArrayList<>();
        for (Repository repo : featuresService.listRepositories()) {
            for (Feature f : repo.getFeatures()) {
                if (f.getName().equals(feature)) {
                    versions.add(f.getVersion());
                }
            }
        }

        return versions;
    }
```

awrb:
The pipeline failed but this probably wasn't caused by my changes? Is this a known instability?
I see something similar here: https://github.com/apache/karaf/runs/8186042837

jbonofre:
Yes there are couple of flaky tests.

jbonofre (approved):
Maybe worth (later) to add a `status` operation on the FeatureMBean.

PR #1626, opened 2022-09-23
[KARAF-7068] add instance:package command to compress an instance into zip/tar.gz

awrb:
Not sure if this one was up for grabs since @jbonofre is already assigned on Jira?


jbonofre:
@awrb good one ! Let me review it, but it's exactly what I had in mind. Thanks !

assemblies/features/standard/src/main/feature/feature.xml
```diff
@@ -630,6 +630,7 @@ update = manager
     </feature>
 
     <feature name="instance" description="Provide Instance support" version="${project.version}">
+        <bundle dependency="true">mvn:org.apache.commons/commons-compress/${commons-compress.version}</bundle>
```
jbonofre, line 633: I would prefer to avoid `commons-compress` dependency, and directly use JDK `Zip*` classes.

I can update the PR if you want.

jbonofre (changes requested):

assemblies/features/standard/src/main/feature/feature.xml
awrb: I'm happy for you to update it :+1:  Is there a tar option in JDK? There is `java.util.zip.GZIPOutputStream` and `java.util.zip.ZipOutputStream` but I don't know anything native for tarballs?

jbonofre: There's no `tar` direct support in JDK. But a `tar` file is a uncompress kind of concat files with some headers. It's possible to create directly this file in Java, respecting the headers, offset, etc. It's basically what `commons-compress` is doing in the `tar` archiver.
As `instance:package` is an extra feature, I think it's fine to do just a zip archive (using `Zip*` JDK level classes).

jbonofre: Thoughts ?

jbonofre: @awrb any update about removing `commons-compress` and using only zip ? I can do that if you prefer. Thanks !

awrb: I will modify the .zip today but not sure about .tar.gz. I'll look into the code of `commons-compress` how to do it. I'll ask you to take over if it takes too long.

jbonofre: I think it's fine to only generate zip archive (not need to create tar.gz).

awrb: I removed `commons-compress` dependency and just left the zip option. Sadly maintaining file permissions (e.g. 0755 in bin/karaf) does not seem to work. Seems to be a bug/limitation in JDK:
https://bugs.openjdk.org/browse/JDK-6194856
https://stackoverflow.com/questions/10699334/change-permission-of-a-zipentry

jbonofre: Yes, it's normal in a zip file. Let's start with this first functionnality, we will improve on user requests.

jbonofre (approved):
LGTM ! Thanks !

I will simply refactore a little bit to add `package` operation in the `InstanceService` and use this in the `package` command and `InstanceMBean` (to be consistent).

PR #1627, opened 2022-09-24
[KARAF-6321] - add configuration options to protect against CTRL-D/shell:logout exits

awrb:


shell/core/src/main/java/org/apache/karaf/shell/impl/console/ConsoleSessionImpl.java
```diff
@@ -451,7 +449,12 @@ private CharSequence readCommand(AtomicBoolean reading) throws UserInterruptExce
                 command = reader.getBuffer().toString();
             }
         } catch (EndOfFileException e) {
-            command = null;
+            boolean disableEofExit = (boolean) session.get(Session.DISABLE_EOF_EXIT);
+            if (disableEofExit) {
+                command = "";
```
awrb, line 454: I tested it manually and it works fine but I couldn't figure out a way to write a test for this (e.g. how to send CTRL-D to Karaf in a pax exam test, or some other way)

jbonofre (commented):
Overall, it looks good (I will take a deeper look). Do you mind to update the document (in the `manual` module) ?

awrb:
Sure @jbonofre I will do it in a couple of hours. Actually some of my other pull requests might also need a change in the `manual`. 
Do you have any ideas how to code a test for the diasabled EOF? 


> Overall, it looks good (I will take a deeper look). Do you mind to update the document (in the `manual` module) ?

awrb:
I updated `console.adoc` and `remote.adoc` in `manual`.

shell/core/src/main/java/org/apache/karaf/shell/impl/console/ConsoleSessionImpl.java
jbonofre: That's actually pretty hard in pax-exam as it controls the lifecycle. Manual test is fine and enough IMHO.

jbonofre (approved):

jbonofre (approved):
Thanks for the documentation update! I will squash and merge.

PR #1628, opened 2022-09-26
[KARAF-1381] - store karaf command history separately for each instance

awrb:
With this change (unless overriden via system properties for a given instance): 
- `/home/user/.karaf/karaf.history` becomes `/home/user/.karaf/karaf.history.root`,
- then each instance has its own history stored in `/home/user/.karaf/karaf.history.{instanceName}`.

### DIFF
--- a/shell/core/src/main/java/org/apache/karaf/shell/impl/console/ConsoleSessionImpl.java
+++ b/shell/core/src/main/java/org/apache/karaf/shell/impl/console/ConsoleSessionImpl.java
@@ -299,7 +299,12 @@ public class ConsoleSessionImpl implements Session {
      * @return the history file
      */
     protected Path getHistoryFile() {
-        String defaultHistoryPath = new File(System.getProperty("user.home"), ".karaf/karaf.history").toString();
+        String instanceName = System.getProperty("karaf.name", "root");
+        String historyFilePath = ".karaf/karaf.history";
+        if (!instanceName.equals("root")) {
+            historyFilePath += ("." + instanceName);
+        }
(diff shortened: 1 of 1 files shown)


shell/core/src/main/java/org/apache/karaf/shell/impl/console/ConsoleSessionImpl.java
```diff
@@ -299,7 +299,9 @@ public Object evaluate(LineReader reader, ParsedLine line, String func) throws E
      * @return the history file
      */
     protected Path getHistoryFile() {
-        String defaultHistoryPath = new File(System.getProperty("user.home"), ".karaf/karaf.history").toString();
+        String instanceName = System.getProperty("karaf.name", "root");
```
awrb, line 302: Might want to keep `karaf.history` filename for the root instance for retrocompatibility? If so there needs to be a condition here.

jbonofre: Yes, if instance name is `root`, it would be great to keep `karaf.history` for the name (for backward compatibility).

jbonofre (changes requested):

shell/core/src/main/java/org/apache/karaf/shell/impl/console/ConsoleSessionImpl.java
awrb: Done, it will stay `karaf.history` now!

jbonofre (approved):

PR #1638, opened 2022-10-08
[KARAF-6769] - fix a bug where command options are not censored when using their aliases

awrb:


jbonofre (approved):
Good catch ! Thanks !

oscerd (approved):

PR #1649, opened 2022-10-12
[KARAF-6319] - add <firstStageBootFeatures> option to assembly mojo

awrb:


manual/src/main/asciidoc/developer-guide/karaf-maven-plugin.adoc
```diff
@@ -471,6 +471,9 @@ The `karaf:assembly` goal creates a Karaf instance (assembly) filesystem using t
                     <installedBundles>
                         <installedBundle>${additional.bundle}</installedBundle>
                     </installedBundles>
+                    <prerequisiteFeatures>
+                        <feature>ssh</feature>
```
jbonofre, line 475: Are you sure about `prerequisiteFeature` name ? Maybe it's a bit confusing for users as `prerequisite` is also a flag in features repository XML.
That's why my first idea was to use `firstStageBootFeature`, to represent it's a boot feature but in "first" stage.

jbonofre (changes requested):
It's more a question about naming for you. The change looks good anyway.

Thanks !

manual/src/main/asciidoc/developer-guide/karaf-maven-plugin.adoc
awrb: I used this name because it's already used in `GenerateDescriptorMojo`:
```
    /**
     * Name of features which are prerequisites (they still need to be defined separately).
     */
    @Parameter
    private List<String> prerequisiteFeatures = new ArrayList<>();
```
also in provisioning.adoc:
```
===== Feature prerequisites

A prerequisite feature is a special kind of dependency. If you add the `prerequisite` attribute to dependant feature tag then it will force installation and also activation of bundles in the dependant feature before the installation of the actual feature. This may be handy in the case that bundles enlisted in a given feature are not using pre installed URLs such as `wrap` or `war`.

----
<feature name="foo" version="1.0.0">
    <feature prerequisite="true">bar</feature>
    ...
</feature>
----
```
so that was my thought process. I am happy to change it though, `firstStageBootFeatures` is nice as it contains the stage name.

jbonofre: Yes, but the semantic is different: `prerequisite` is a flag in the feature repository XML to define a feature should be fully installed before the current feature. In the case of boot features, it's different because:
1. it's in `etc/org.apache.karaf.features.cfg` not in the features repository XML (so global)
2. it's not really prerequisite but more staging (a feature in first stage could have a prerequisite defined)

If you don't mind, I would prefer to use `firstStageBootFeatures`.

Thoughts ?

awrb: Good points! I agree. I will change the name a bit later today and ping you.

jbonofre: Thank you so much ! Much appreciated !

jbonofre:
@awrb Great ! Many thanks ! I will do a new pass today.

jbonofre (approved):

PR #1657, opened 2022-10-13
Add note about possible use of env variables

splatch:
I found users surprised how environment variables are working with configurations. Effectively in karaf universe you need to have a configuration file first in order to override it values with environment variables. This small tweak adds explicit note for users so they do not run into wrong assumptions.

### DIFF
--- a/manual/src/main/asciidoc/user-guide/configuration.adoc
+++ b/manual/src/main/asciidoc/user-guide/configuration.adoc
@@ -38,6 +38,13 @@ export ORG_APACHE_KARAF_SHELL_SSHPORT=8102
 
 You can override any configuration property using this mechanism.
 
+[NOTE]
+====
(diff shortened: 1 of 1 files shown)


mfrazier63 (changes requested):
Recommend slight change to wording:

"requires the OSGi configuration file (.cfg) to exist"....

"in the configuration file".....

splatch:
@mfrazier63 you're right that in most of cases it will be a cfg file, however configuration processor which does the property matching works at `ConfigurationAdmin` level which might be populated by other mechanisms. Yet, to be fair I don't know any other than one you mentioned.

mfrazier63:
@splatch ok, no big deal if it needs to be more generic. Just wanted to make sure things were clear.

jbonofre (approved):

PR #1662, opened 2022-10-14
[KARAF-1717] - add markdown format option to commands-generate-help goal

awrb:


awrb:
It's pretty tricky to write a test for the mojo, I tried using maven-plugin-testing-harness but couldn't get it to work (help with that would be welcome). Maybe a new example that shows how to use `commands-generate-help` in examples/karaf-maven-example would be good?

jbonofre:
@awrb Thanks for this new contribution. To test a plugin, you can use an integration test (it) with the maven invoker. It's what I did here: https://github.com/apache/karaf/tree/main/tooling/karaf-maven-plugin/src/it/test-dockerfile 

Basically you create a concrete maven project, execute build and verify the result. You can do the same to verify the markdown result.

awrb:
@jbonofre done. It would be good to also add tests for other formats but probably in another PR. Also what do you think of migrating verify.bsh sciripts to Groovy? I think it's more modern.

jbonofre:
@awrb thanks. Yes we can do other PRs for other formats. No problem to use Groovy.

jbonofre (approved):

PR #1677, opened 2022-10-25
[KARAF-6017] - added cache facade

awrb:
Draft phase (it works but it's not a proper component yet).

I'm opening a pr already so it can be reviewed before putting more work into it.

Some features:
- it uses ehcache with JSR 107 API
- the api does not depend on ehcache, so other providers could be added
- possible to load ehcache configuration from xml
- some karaf commands for debugging/playing around with the caches

Some open points:
- how should the interface (`CacheService`) look like
- is anything important not possible because of the facade (see the docs: https://www.ehcache.org/documentation/3.0/index.html)

### DIFF
--- a/cache/api/src/main/java/org/apache/karaf/cache/api/CacheService.java
+++ b/cache/api/src/main/java/org/apache/karaf/cache/api/CacheService.java
@@ -1,0 +1,89 @@
+/**
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.karaf.cache.api;
+
+import javax.cache.Cache;
+import javax.cache.configuration.Configuration;
+import java.net.URL;
+import java.util.List;
+
+/**
+ * A simple caching facade that lets use caching using JSR 107 APIs.
+ */
+public interface CacheService {
+
+    /**
+     * Creates a new cache.
+     * @param name name of the cache to create
+     * @param configuration cache configuration, such as its key and value type.
+     *                      Specific configurations (such as max size and caching duration)
+     *                      will depend on the cache provider.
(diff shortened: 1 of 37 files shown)


cache/api/src/main/java/org/apache/karaf/cache/api/CacheService.java
```diff
@@ -0,0 +1,17 @@
+package org.apache.karaf.cache.api;
```
jbonofre, line 1: Missing ASF header here.

jbonofre (commented):
That's a good start. I will come back to you with more detailed feedback (busy at EclipseCon this week).

cache/api/src/main/java/org/apache/karaf/cache/api/CacheService.java
awrb: It's missing in other places too (still a draft), will fix everything once we know how the interface should look like.

cache/api/src/main/java/org/apache/karaf/cache/api/CacheService.java
```diff
@@ -0,0 +1,17 @@
...
+    <K, V> void createCache(String name, Configuration<K, V> configuration);
+
+    <K, V> V get(String name, K key);
+
+    <K, V> void put(String name, K key, V value);
+
+    void invalidateCache(String name);
+
+    List<String> listCaches();
+}
```
awrb, line 17: A method to get the entire JSR107 `Cache` (other than get/put) might be useful too.

cache/commands/src/main/java/org/apache/karaf/cache/core/commands/Create.java
```diff
@@ -0,0 +1,39 @@
...
+    @Argument(index = 2, required = true)
+    String valueType;
+    @Reference
+    private CacheService cacheService;
+
+    @Override
+    public Object execute() throws Exception {
+        Class keyClass = Class.forName(keyType);
+        Class valueClass = Class.forName(valueType);
+        Configuration<?, ?> configuration = new MutableConfiguration<>()
```
awrb, line 32: This is just hard-coded for now but I'll extend this (add arguments to specify the cache size, or the xml configuration path, probably make the key/value type resolution smarter, so something like `cache:create myCache long string <other args>` also works).

cache/commands/src/main/java/org/apache/karaf/cache/core/commands/Invalidate.java
```diff
@@ -0,0 +1,28 @@
...
+    CacheService cacheService;
+
+    @Argument(index = 0, required = true)
+    @Completion(CacheNameCompleter.class)
+    String cacheName;
+
+    @Override
+    public Object execute() throws Exception {
+        cacheService.invalidateCache(cacheName);
+        return cacheService + " invalidated";
```
awrb, line 26: Should be `cacheName + " invalidated";`

awrb:
@jbonofre I made a proper pull request now (added docs, example, tests etc). I'm happy to introduce any changes if needed.

jbonofre:
This PR is stale and old.

PR #1678, opened 2022-11-03
[KARAF-6697] Fix threads leak in karaf-maven-plugin verify goal

wborn:
Since we now verify ~400 features in the openHAB Addons repo, the number of threads being leaked by the verify goal causes [GitHub Actions builds to run out of memory](https://github.com/wborn/openhab-addons/actions/runs/3350846650/jobs/5554259431).

I did some investigation and found that the thread leak is caused by the VerifyMojo itself.
For some reason the threads created by the executor are never garbage collected.

---

The goal leaks 8 threads each time it is executed. So if you verify 400 features in a build it causes 3200 threads to be leaked:

before

---

When the executor is shutdown the verify goal no longer leaks threads. :slightly_smiling_face: 

after

---

It would also be nice when the fix is cherry-picked to 4.3.x.

### DIFF
--- a/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/VerifyMojo.java
+++ b/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/VerifyMojo.java
@@ -48,8 +48,6 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.BiFunction;
-import java.util.function.Function;
-import java.util.function.Supplier;
 import java.util.jar.Attributes;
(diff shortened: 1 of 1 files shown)


tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/VerifyMojo.java
```diff
@@ -48,8 +48,6 @@
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.BiFunction;
-import java.util.function.Function;
-import java.util.function.Supplier;
```
wborn, line 52: These are unused imports.

oscerd (approved):

jbonofre (approved):

PR #1713, opened 2023-02-07
KARAF-5421 - add user friendly error when installing non OSGi bundle …

awrb:
…via bundle:install

### DIFF
--- a/bundle/core/src/main/java/org/apache/karaf/bundle/command/Install.java
+++ b/bundle/core/src/main/java/org/apache/karaf/bundle/command/Install.java
@@ -49,13 +49,13 @@ public class Install implements Action {
 
     @Option(name = "-l", aliases={"--start-level"}, description="Sets the start level of the bundles", required = false, multiValued = false)
     Integer level;
-    
+
     @Option(name = "--force", aliases = {"-f"}, description = "Forces the command to execute", required = false, multiValued = false)
     boolean force;
 
     @Option(name = "--r3-bundles", description = "Allow OSGi R3 bundles")
     boolean allowR3;
-    
+
     @Reference
     Session session;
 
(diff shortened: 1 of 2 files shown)


jbonofre:
Did you check if you directly install a blueprint xml ? The blueprint xml doesn't contain manifest so I think your change will block it.

awrb:
Good point @jbonofre, I will check in the evening. If it doesn't work then I'll add a test since the current ones did not detect it if it broke. Although the previous code looks like it wouldn't work too, because it was either manifest version == "2" or exception (unless --r3-bundles is passed).

jbonofre:
Yes I think blueprint installer is creating a bundle on the fly. So that's why it works.

awrb:
> Yes I think blueprint installer is creating a bundle on the fly. So that's why it works.

The docs say it does https://karaf.apache.org/manual/latest/#_blueprint_deployer
```
The generated OSGi MANIFEST will contain the following headers:

Manifest-Version: 2
Bundle-SymbolicName: [name of the file]
Bundle-Version: [version of the file]
Import-Package: [required packages]
DynamicImport-Package: *
```
but I am not sure when this happens compared to bundle:install.

awrb:
@jbonofre I just tried manually and it looks like the blueprint bundles install fine, e.g. the bundles of mvn:org.apache.karaf.examples/karaf-blueprint-example-* (as confirmed by karaf/itests/test/src/test/java/org/apache/karaf/itests/examples/BlueprintExampleTest.java passing in the pipeline).

But they are already bundles because of `<packaging>bundle</packaging>` and `maven-bundle-plugin`. Blueprint just handles service dynamics. Am I missing something?

jbonofre:
@awrb that's blueprint bundles, I mean that it's possible to directly deploy blueprint XML in Karaf:

```
bundle:install blueprint:mvn:...
```

I will test myself later today or tomorrow.

awrb:
image
after `bundle:install -s blueprint:file:...` it generates Bundle-ManifestVersion automatically for me. But I'm just testing one use case so if you can test too that's great.

jbonofre (approved):

PR #1716, opened 2023-03-13
[KARAF-7678] Fix ShellTable multiline clipping

CMoH:
Account for multi-line cell contents when calculating cell width and clipping the text to maxSize. The cell contents might be either wrapped or originally split by the caller with \n characters

### DIFF
--- a/shell/core/src/main/java/org/apache/karaf/shell/support/table/Col.java
+++ b/shell/core/src/main/java/org/apache/karaf/shell/support/table/Col.java
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import org.apache.karaf.shell.support.ansi.SimpleAnsi;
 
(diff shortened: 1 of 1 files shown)


oscerd (approved):

shell/core/src/main/java/org/apache/karaf/shell/support/table/Col.java
```diff
@@ -148,8 +149,14 @@ String format(Object cellData) {
...
-        String finalContent = cut(fullContent, getClippedSize(fullContent.length()));
-        updateSize(finalContent.length());
+        if (wrap && size < fullContent.length()) {
+            // make sure splitLines will have an estimate cell size if wrap is true
+            updateSize(fullContent.length());
+        }
+        List<String> lines = splitLines(fullContent);
+        int maxLineSize = lines.stream().mapToInt(String::length).max().getAsInt(); // at least one line exists due to test above
+        updateSize(maxLineSize); // calls getClippedSize()
+        String finalContent = lines.stream().map(line -> cut(line, getClippedSize(line.length()))).collect(Collectors.joining("\n"));
```
wborn, line 159: Will this work on Windows which uses a different line separator?
E.g. using `System.lineSeparator()` would be based on the OS.

CMoH: I think in this particular case I think it will work, since `\n` is included in the Windows line separator, `\n\r`. Also, the contents of `splitLines()` is the verbatim top-section of the `getContents()` method, so at least I don't expect the behaviour to change after this commit.

However, the comment is valid in principle. As you mentioned in https://github.com/apache/karaf/pull/1715#discussion_r1139063272, this seems to be a style issue across the entire code base. If so, would it be worth adressing them all as a separate issue?

jbonofre (approved):

shell/core/src/main/java/org/apache/karaf/shell/support/table/Col.java
jbonofre: I would address the split line separator in another PR. Anyway I don't think it's a big deal :)

PR #1724, opened 2023-06-14
Ensure jre and eecap config.properties properties can be computed for modern jvms

rmannibucau:
Mainly to show how to not have to update properties each 6 months (see #1723 for context)

### DIFF
--- a/util/src/main/java/org/apache/karaf/util/config/PropertiesLoader.java
+++ b/util/src/main/java/org/apache/karaf/util/config/PropertiesLoader.java
@@ -18,17 +18,19 @@
  */
 package org.apache.karaf.util.config;
 
+import org.apache.felix.utils.properties.InterpolationHelper;
+import org.apache.felix.utils.properties.Properties;
+
 import java.io.File;
 import java.io.FileNotFoundException;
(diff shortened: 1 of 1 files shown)


fpapon (approved):
LGTM
Thanks!

jbonofre:
NB: the PR doesn't prevent the update in `etc/jre.properties` but help to deal with ee caps.

mattrpav:
@rmannibucau shouldn't the eecap-* entries the be able to be removed from the etc/config.properties? 

This approach looks way better!

rmannibucau:
@mattrpav the eecap and jre props defaulting to a previous value (or just appending their number) can be removed until java 9 (to keep). One option which is not bad is to decide to define all LTS and skip others, this enables to not loop too much but also to work most of the time without any explicit config so think it can be a wise compromise.

jbonofre:
@mattrpav we can't remove the old entries for some version of JDKs. However it's possible for Karaf 4.5.x.

jbonofre:
I'm testing/reviewing this PR. Thanks !

jbonofre (approved):

PR #1745, opened 2023-07-29
KARAF-7732 - Upgrade to Felix Configurator 1.0.18

jbonofre:


oscerd (approved):

jbonofre:
There's something weird in felix configurator 1.0.18: it imports org.apache.felix.cm.json.io package but this package is not exported by the felix cm json bundle (at least in 1.0.x). I'm checking.

PR #1771, opened 2023-09-22
[KARAF-7754] Require Java 11 at runtime/build time

rovarga:
This switches enforcer to require at least Java 11 at build-timetime and karaf's launch enforcement to require Java 11+ at run-time.


jbonofre:
Maybe we should change the `javase` version as well (in the `karaf-maven-plugin` and so).

rovarga:
> Maybe we should change the `javase` version as well (in the `karaf-maven-plugin` and so).

Ah yes, I did not get that far. Since that is a run-time requirement, I am on the fence on that: it definitely has to be done at the latest when we switch to java.compiler.release to 11.

OTOH this PR unblocks things like Pax Web upgrade, so I'd like to keep its scope pretty much minimal.

wborn:
There is also some logic in the [karaf](https://github.com/apache/karaf/blob/856288142047003bef0d395ef175fc55bc47a2b6/assemblies/features/base/src/main/filtered-resources/resources/bin/karaf#L303-L358) and [karaf.bat](https://github.com/apache/karaf/blob/856288142047003bef0d395ef175fc55bc47a2b6/assemblies/features/base/src/main/filtered-resources/resources/bin/karaf.bat#L419-L470) scripts that can be simplified when Java 9+ is required.

rovarga:
> There is also some logic in the [karaf](https://github.com/apache/karaf/blob/856288142047003bef0d395ef175fc55bc47a2b6/assemblies/features/base/src/main/filtered-resources/resources/bin/karaf#L303-L358) and [karaf.bat](https://github.com/apache/karaf/blob/856288142047003bef0d395ef175fc55bc47a2b6/assemblies/features/base/src/main/filtered-resources/resources/bin/karaf.bat#L419-L470) scripts that can be simplified when Java 9+ is required.

You are right. As mentioned in the issue, we need to take a staggered approach here. I will split the effort accordingly.

PR #1781, opened 2023-09-25
[KARAF-7754] Target Java 11

rovarga:

### DIFF
--- a/management/server/pom.xml
+++ b/management/server/pom.xml
@@ -125,6 +125,26 @@
         </resources>
         <plugins>
             <plugin>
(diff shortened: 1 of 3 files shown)


oscerd (approved):

jbonofre:
It doesn't work for verify goal with this change. The verify mojo should be updated

jbonofre (changes requested):
We need a change on the karaf-maven-plugin:verify mojo to be able to deal with Java 11 kind packages.

rovarga:
Ah, yeah, I think this needs a lot more work. I will open a new PR once we are ready to take this step.

PR #1784, opened 2023-10-02
[KARAF-7768]: Provide SPI Fly from framework

rovarga:


assemblies/features/framework/src/main/feature/feature.xml
```diff
@@ -45,6 +45,20 @@
...
         <!-- features service -->
         <bundle start-level="15">mvn:org.apache.karaf.features/org.apache.karaf.features.core/${project.version}</bundle>
+
+        <!--
+            ASM/SPI Fly integration.
+
+            spifly.dynamic.bundle and not spifly.dynamic.framework.extension, in order to lower ASM library duplication: other projects
+            are using them. A more packaging-intensive operation would be to package these 6 bundles as 1-2 framework extensions.
+          -->
+        <bundle dependency="false" start-level="20">mvn:org.ow2.asm/asm/${asm.version}</bundle>
```
jbonofre, line 55: do we really think it's a good idea to provide spifly at framework level ? Just thinking :)

rovarga: Yeah, that got me thinking as well, simply because at that point we really are guided towards spifly.dynamic.framework.extension, which is a heavyweight.

Having thought about this quite a bit a more, I believe want to do something else here.

rovarga:
This PR does the wrong thing, I will raise another one.

PR #1789, opened 2023-11-05
avoid long evaluation chain for JRE property

ecki:


jbonofre (approved):
It looks good, even if I'm not sure it helps readiness ;)

PR #1804, opened 2023-12-14
KARAF-6210: extract core config commands to its own bundle

RangerRick:
This PR performs a simple split of `org.apache.karaf.config.core` into `org.apache.karaf.config.core` and `org.apache.karaf.config.command` which resolves the problem of relying on the shell api in the main config core.


jbonofre:
@RangerRick Thanks ! That's a good idea, good catch. I will review and test the standard feature (four eyes verification :) ). I will probably include this in Karaf 4.5.0 and 4.4.5.

RangerRick:
I'm testing this out with a patch against 4.3.10 right now (which is what we still currently rely on, haven't figured out why we're having issues moving to 4.4 yet)

jbonofre:
@RangerRick I just wonder if the root cause of the problem is really the shell in config. It sounds more like a race condition for me. So, this is not a fix but more a workaround imho.

RangerRick:
@jbonofre Ah, hrm, realizing that `org.apache.karaf.config.command` doesn't get pulled in by the standard featureset, since command.core is specified as a framework/startup bundle.
What would be the correct place to add it to make sure it shows up properly? Somewhere in the framework feature file?

RangerRick:
> @RangerRick I just wonder if the root cause of the problem is really the shell in config. It sounds more like a race condition for me. So, this is not a fix but more a workaround imho.

@jbonofre I'm really not sure. It is definitely a race condition, but it's not clear what the "right" answer, a bunch of this stuff is so intertwined. I've been fighting this for a couple of weeks.

We have a fun thing where we implement our own config impl, so we're running into this as a part of updating our Karaf version. I suspect we're more likely than most to hit issues because our Karaf distribution looks less "standard".

RangerRick:
erp, used to my internal workflow, I should probably not be force-pushing test changes here, sorry about that :)

rovarga:
> @jbonofre Ah, hrm, realizing that `org.apache.karaf.config.command` doesn't get pulled in by the standard featureset, since command.core is specified as a framework/startup bundle. What would be the correct place to add it to make sure it shows up properly? Somewhere in the framework feature file?

I wonder whether [KARAF-7775](https://issues.apache.org/jira/browse/KARAF-7775) is part of the picture here

RangerRick:
@rovarga Ooh, yes! I have noticed that when the `KARAF-6210` issue happens, trying to do `config:edit` gets eaten and `scr:config` is the only config command available in my shell.

rovarga:
So as far as my understanding goes, we want to structure this just as features -- i.e. there is feature.core and features.command -- and the latter is installed conditionally based on the shell feature (IIRC, but a quick grep should confirm that).

RangerRick:
Also now that I try a fresh build, I find it's failing because there's some introspection from commands into `org.apache.karaf.config.core.impl` so maybe this needs a bit more refactoring than I'd hoped.

I'll play with this a bit and get back to you...

rovarga:
I mean this bit: https://github.com/apache/karaf/blob/7761d9be81e0fa5f26882842666cff72b5d2a257/assemblies/features/standard/src/main/feature/feature.xml#L125-L157

jbonofre:
@RangerRick that's why I was suspected.

@rovarga possible, I will take a new look

RangerRick:
> So as far as my understanding goes, we want to structure this just as features -- i.e. there is feature.core and features.command -- and the latter is installed conditionally based on the shell feature (IIRC, but a quick grep should confirm that).

So I'm struggling to understand what feature to put that _in_ though. My thought was to do this:

```diff
diff --git a/assemblies/features/standard/src/main/feature/feature.xml b/assemblies/features/standard/src/main/feature/feature.xml
index 85e9e04ce0..96c8fcd180 100644
--- a/assemblies/features/standard/src/main/feature/feature.xml
+++ b/assemblies/features/standard/src/main/feature/feature.xml
@@ -527,2 +527,7 @@ watch = admin
     <feature name="config" description="Provide OSGi ConfigAdmin support" version="${project.version}">
+        <bundle start-level="11">mvn:org.apache.karaf.config/org.apache.karaf.config.core/${project.version}</bundle>
+        <conditional>
+            <condition>shell</condition>
+            <bundle start-level="30">mvn:org.apache.karaf.config/org.apache.karaf.config.command/${project.version}</bundle>
+        </conditional>
         <conditional>
diff --git a/itests/test/src/test/filtered-resources/etc/feature.xml b/itests/test/src/test/filtered-resources/etc/feature.xml
index 7926d51995..a334e9d2c7 100644
--- a/itests/test/src/test/filtered-resources/etc/feature.xml
+++ b/itests/test/src/test/filtered-resources/etc/feature.xml
@@ -360,3 +360,6 @@
         <bundle start-level="11" start="true">mvn:org.apache.karaf.config/org.apache.karaf.config.core/${project.version}</bundle>
-        <bundle start-level="30" start="true">mvn:org.apache.karaf.config/org.apache.karaf.config.command/${project.version}</bundle>
+        <conditional>
+            <condition>shell</condition>
+            <bundle start-level="30" start="true">mvn:org.apache.karaf.config/org.apache.karaf.config.command/${project.version}</bundle>
+        </conditional>
         <conditional>
```

...does that seem reasonable?

jbonofre:
@RangerRick your change on the feature doesn't validate. It should be using conditional. If you don't mind, I will do the change/fix for you.

RangerRick:
Yeah I got stuck working on other things before the holidays, I realized that as soon as I saw the build failures. Was gonna take another look in the new year but if it's a quick fix I'd appreciate it.

jbonofre:
@RangerRick I found other issue with your bundles: the import required impl package which is private, it should not export utils, etc. I propose to close this one and do a clean new one.

jbonofre:
Superseded by #1814

RangerRick:
@jbonofre Awesome, thanks! I got backed up over the holidays and never had a chance to look further but I was going down the same path (although not as in-depth as what you ended up doing) before I set it aside.

PR #1809, opened 2023-12-16
[KARAF-7544] Clean up Jetty/servlet-api use

rovarga:
Centralize Jetty version in a property and clean up references to servlet-api, making future upgrades easier.


oscerd (approved):

grgrzybek (approved):

jbonofre:
I will do a manual merge to test and squash.

jbonofre:
FYI: it's what we had before but we changed that as Jetty and servlet api is provided by Pax Web. However, for backward compatibility we add a meta feature. In Karaf 4.5.x, we can completely remove `jetty` feature from Karaf and use `pax-web-jettty` from pax web instead.

PR #1816, opened 2024-01-13
[KARAF-7768] Remove org/apache/karaf/specs

rovarga:
specs.locator and related logic just dances around the need to integrate with ServiceLoader. In OSGi world that's properly done with adding the appropriate metadata and therefore this magic should not be used.


jbonofre:
Superseeded by #1958

PR #1822, opened 2024-01-26
KARAF-7804 - provide unique servlet-name when registering http proxy servlets

awrb:


awrb:
In the Jira ticket I explain the rationale behind this. For the servlet-name I think simply the proxy URL could be used (without slash), e.g. when proxying to `/external-api` the unique proxy servlet name is `external-api`.

jbonofre (approved):

PR #1861, opened 2024-07-24
[KARAF-7854] Upgrade to jackson 2.17.2

geichelberger:
Upgrades the following dependencies:
- snakeyaml from 1.33 to 2.2
- jackson from 2.14.1 to 2.17.2


jbonofre:
I suspect the feature might have issue with snakeyml update. Let me double check.

jbonofre:
I tested it again and it's ok. Thanks !

jbonofre (approved):

PR #1863, opened 2024-09-20
KARAF-5014: consider first group role in users.properties and ignore empty roles

stataru8:
This issue is quite old and only affects those who directly modify the `users` file. I'm not sure if it is still relevant, or if the current behaviour should be considered acceptable.

Current behaviour:
- the first role in a group is ignored
- when a group is created via `jaas` commands, the role `group` is added by default

Proposal:
 - consider the first role in a group
 - when a group is created using `jaas` commands, add an empty role by default
 - ignore empty roles in both users and groups


AndreVirtimo:
Hi @stataru8 , thank you for picking up this issue.

Why do you propose "when a group is created using jaas commands, add an empty role by default"? Is this because there is no role which could be assigned during creation? Is it not possible to have an empty group?

Regards
Andre

jbonofre (approved):

jaas/modules/src/main/java/org/apache/karaf/jaas/modules/properties/PropertiesBackingEngine.java
```diff
@@ -52,14 +52,13 @@ public void addUser(String username, String password) {
         if (username.startsWith(GROUP_PREFIX))
             throw new IllegalArgumentException("Prefix not permitted: " + GROUP_PREFIX);
 
-        addUserInternal(username, password);
+        addUserInternal(username, encryptionSupport.encrypt(password));
```
jbonofre, line 55: Why forcing encryption here ? It's an optional feature.

jaas/modules/src/main/java/org/apache/karaf/jaas/modules/properties/PropertiesBackingEngine.java
```diff
@@ -157,22 +159,37 @@ private List<RolePrincipal> listRoles(String name) {
...
+
     @Override
     public void addRole(String username, String role) {
         String userInfos = users.get(username);
         if (userInfos != null) {
-            for (RolePrincipal rp : listRoles(username)) {
-                if (role.equals(rp.getName())) {
-                    return; 
+
+            // groups don't have password and empty should be ignored
```
jbonofre, line 174: A group can be empty (no role, no group, no user).

jbonofre (changes requested):

jaas/modules/src/main/java/org/apache/karaf/jaas/modules/properties/PropertiesBackingEngine.java
stataru8: I just moved this call from its original location in `addUserInternal`:
https://github.com/apache/karaf/blob/e9b9c973569596c8931cc4e8f7d62744d9c3ede5/jaas/modules/src/main/java/org/apache/karaf/jaas/modules/properties/PropertiesBackingEngine.java#L62

The call is only needed when adding a user and shouldn't be made when adding a group. There is the risk of encrypting `""`, which with the defaults, results in `{CRYPT}ABC...{CRYPT}`. After `jaas:group-add karaf newGrup`, 
`jaas:user-list` will return 
```
User Name | Group   | Role
----------+---------+-------------------------------------------------------------------------------
karaf     | newGrup | {CRYPT}ABC...{CRYPT}
```

Maybe at this point, we should create another `addUserInternal` with just the `username` as an argument: `private void addUserInternal(String username)`, or another function just for groups...

jaas/modules/src/main/java/org/apache/karaf/jaas/modules/properties/PropertiesBackingEngine.java
stataru8: My original comment was a bit misleading, what I meant is:
- If a **user** info is empty, we shouldn't replace `""` by `role`, or else `role` becomes the user's password.
- If a **group** info is empty, we should replace `""` by `role`.

stataru8:
Hello @AndreVirtimo, as far as I know, we have two commands that allow us to create groups:
- `jaas:group-add user myGroup1`
- `jaas:group-create myGroup2`

In both cases, there is no mandatory argument for the role. Today, the role/password `group` is added by default. The proposal is to add `""` instead, leaving the group information empty.

jbonofre:
Sorry for the delay in the review/comment. I want to provide some context.

The idea of the first "role" in a group definition, it's not actually a role but just a "prefix" (`group,`) to indicate that we are in a group.

That's why be default, in `etc/users.properties` you have:

```
_g_\:admingroup = group,admin,manager,viewer,systembundles,ssh
```

You can see that `group` here is not actually a role in `admingroup` but a "prefix flag".

I agree it's confusing (including the documentation), but if we "change" to remove the prefix, it means that `group` would be considered as a role in `admingroup`.

jbonofre (changes requested):
Can you please update default `etc/users.properties` to remove `group` as it's not actually a role ?

stataru8:
I just realized that `group` is also used as a "prefix flag" in `keys.properties`. I wonder if the changes need to be propagated in this login module as well:
https://github.com/apache/karaf/blob/a8d990102fd5181bcb8e23ef8fe06a5782db9d0a/jaas/modules/src/main/java/org/apache/karaf/jaas/modules/publickey/PublickeyBackingEngine.java#L273

Let me know if is that's too much for this PR or out of scope. 

If all these changes are too risky, just making the documentation clearer about the "prefix flag" is also an option...

jbonofre:
@stataru8 yes, historically, we use this "flag" in all group/role definition (users, keys, ...). So, we can consider this as a breaking change for users (if we remove the group prefix).
I wonder what's the best option:
1. Documenting clearly the group prefix (but I wonder if it's still useful to have this prefix)
2. Remove the use of group prefix for Karaf 4.5.0 (as it's a breaking change)
Thoughts ?

AndreVirtimo:
I don't think we need a flag in the value. Because we already have _g_ in the key.

So I am in favor of a change. I think the risk is low. If users don't migrate their configuration, they get an additional role “group”, which shouldn't hurt.

stataru8:
I'll try to commit the changes for `keys.properties` this week or next. Will request a re-review once it's done

stataru8:
I see some duplicate code in `jaas/modules/src/main/java/org/apache/karaf/jaas/modules/properties/PropertiesBackingEngine.java` and `jaas/modules/src/main/java/org/apache/karaf/jaas/modules/publickey/PublickeyBackingEngine.java`. 
Should we take this opportunity to refactor? Is it safe to do so? I was thinking of creating a parent class named `AbstractPropertiesBackingEngine` or `FileBasedBackingEngine`...

jbonofre:
@stataru8 as we target that for 4.5.x, I don't have problem if we include a refactoring (especially a cleanup one 😄 ).

stataru8:
Hello again, one more question, should the PR also include changes to the `.adoc` files?
https://github.com/apache/karaf/blob/ad427cd12543dc78e095bbaa4608d7ca3d5ea4d8/manual/src/main/asciidoc/user-guide/security.adoc?plain=1#L142

jaas/command/src/main/java/org/apache/karaf/jaas/command/ListUsersCommand.java
```diff
@@ -104,6 +105,8 @@ private List<String> displayGroupRoles(BackingEngine engine, String userName, Gr
                 names.add(roleName);
                 table.addRow().addContent(userName, group.getName(), roleName);
             }
+        } else {
+            table.addRow().addContent(userName, group.getName(), "");
```
stataru8, line 109: I don't know if this proposal makes sense. I can revert.

**Setup: user `foo` without any groups or roles declared.**

Current behaviour
```
karaf@trun()> jaas:group-add foo g1
karaf@trun()> jaas:update
karaf@trun()> jaas:realm-manage --realm karaf --module org.apache.karaf.jaas.modules.properties.PropertiesLoginModule
karaf@trun()> jaas:user-list
User Name │ Group      │ Role
──────────┼────────────┼──────────────
foo       │            │
```

Proposal
```
karaf@trun()> jaas:group-add foo g1
karaf@trun()> jaas:update
karaf@trun()> jaas:realm-manage --realm karaf --module org.apache.karaf.jaas.modules.properties.PropertiesLoginModule
karaf@trun()> jaas:user-list
User Name │ Group      │ Role
──────────┼────────────┼──────────────
foo       │ g1         │
```

Proposal - more tests
```
karaf@trun()> jaas:role-add foo r1
karaf@trun()> jaas:update
karaf@trun()> jaas:realm-manage --realm karaf --module org.apache.karaf.jaas.modules.properties.PropertiesLoginModule
karaf@trun()> jaas:user-list
User Name │ Group      │ Role
──────────┼────────────┼──────────────
foo       │ g1         │
foo       │            │ r1
karaf@trun()> jaas:group-role-add g1 r1
karaf@trun()> jaas:update
karaf@trun()> jaas:realm-manage --realm karaf --module org.apache.karaf.jaas.modules.properties.PropertiesLoginModule
karaf@trun()> jaas:user-list
User Name │ Group      │ Role
──────────┼────────────┼──────────────
foo       │ g1         │ r1
```

stataru8: Maybe it makes sense for the last output to be
```
User Name │ Group      │ Role
──────────┼────────────┼──────────────
foo       │ g1         │ r1
foo       │            │ r1
```

stataru8:
PR is ready for review again. 
For the **PublickeyLoginModule**, I was only able to test roles and groups in debug mode using `ssh -p 8101 -i ~/karaf.id_rsa test@localhost`. For some reason, two logins are required: the first login through PublickeyLoginModule succeeds, but then I get prompted for a password (as a string) and since I can't pass the key...
```
C:\Users\stataru\Downloads\SSHTEST\apache-karaf-4.4.6\bin>ssh -p 8101 -i karaf.id_rsa test@localhost
Password authentication
(test@localhost) Password:
```

mattrpav:
This should be karaf 4.5.x (or higher) and clearly called out in a Release Notes / Migration Guide.

Make a security-related change to the file and _presume_ the extra role 'group' would not be an issue is not a good practice.

jbonofre:
I asked the question of breaking change. For now I consider a 4.5.x thing.

mattrpav:
Is the bug here really the JAAS karaf commands / Encryption Support need to handle the 'group' placeholder better?

keys.properties, and ability to support other security providers with a similar file format is useful -- this would allow for sharing of JAAS commands to be able to resolve group/role info for non-local security providers, etc.

For example: LDAP support could be improved to move the group/role mapping _out_ of the LDAP config and simply into a properties file of the same format that maps userDN'groupDN to roles, etc.

I'm trying to understand what the real 'problem' is with the current solution? Perhaps the Karaf JAAS commands simply need to be fixed?

PR #1884, opened 2024-11-12
KARAF-7174: Implement a SSH channel resource cleaner whiteboard

jbonofre:


mattrpav:
This is a great idea and useful feature. Esp for other commands that use connection pools, or other stateful resources.

One suggestion -- perhaps a rename for consistency?

CommandCloseListener?

jbonofre:
Yes,
Good point about the name. My only comment is that (potentially) it could be used for other services than commands. But let's start with shell commands :)

mattrpav:
> Yes, Good point about the name. My only comment is that (potentially) it could be used for other services than commands. But let's start with shell commands :)

How do you see it working for other services? I think things like scr services should use deactivate(). I'm struggling to find a use case where another service would go 'out-of-scope' so to speak similar to how the shell command can timeout.

shell/ssh/src/main/java/org/apache/karaf/shell/ssh/Activator.java
```diff
@@ -189,6 +194,20 @@ protected SshServer createSshServer(SessionFactory sessionFactory) {
...
         server.setSignatureFactories(SshUtils.buildSigAlgorithms(sigAlgorithms));
         server.setShellFactory(new ShellFactoryImpl(sessionFactory));
+        server.addSessionListener(new SessionListener() {
+            @Override
+            public void sessionDisconnect(org.apache.sshd.common.session.Session session, int reason, String msg, String language, boolean initiator) {
+                try {
+                    Collection<ServiceReference<ChannelResourceCleaner>> references = bundleContext.getServiceReferences(ChannelResourceCleaner.class, null);
+                    for (ServiceReference<ChannelResourceCleaner> reference : references) {
+                        ChannelResourceCleaner cleaner = bundleContext.getService(reference);
+                        cleaner.close();
```
mattrpav, line 204: This appears to call .close() on _every_ reference for every session that closes. Is that the correct association? 

Wouldn't we want to close only the instance of the command associated with that session?

mattrpav:
> @mattrpav I see a use case: a command use a service which starts a thread. The service has to be hook with ssh session to stop the thread when the session disconnects. So the service is not a shell command but it has to clean resources sync with ssh session. SCR deactivate is a hook on SCR component lifecycle not the ssh session.

What about simply implementing the JDK's Closable and follow a convention vs marking with a domain-specific interface? 

Seems like (commandInstance instanceof Closeable) would do the trick and we wouldn't need to create a new interface. Thoughts?

shell/ssh/src/main/java/org/apache/karaf/shell/ssh/Activator.java
jbonofre: That's right it's a bit agressive: if we have several ssh clients on the Karaf ssh server, we will close *all* references whatever is the client.
It's pretty hard to "link" the session with the command (or maybe we can use a session property to identify the correct session).

jbonofre:
@mattrpav that's a good idea to leverage `Closeable`. Let me do an experiment.

shell/ssh/src/main/java/org/apache/karaf/shell/ssh/Activator.java
mattrpav: While we're at adding things to the Command/Action session.. this change is needed due to JDK removal of Security Manager: https://issues.apache.org/jira/browse/KARAF-7805

PR #1888, opened 2024-12-19
KARAF-7882: Upgrade to Pax URL 2.6.15

jbonofre:


jbonofre:
I will relaunch Jenkins when Pax URL artifacts will be on Maven Central.

PR #1937, opened 2025-04-24
[KARAF-7950] Use modular OSGi dependencies

rovarga:
osgi-core is non-modular, i.e. it has no Automatic-Module-Name declaration.

This patches switches to using org.osgi.{dto,framework,resource,...}, which are provide this stability.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -41,10 +41,109 @@
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
+
+            <!-- Monolithic OSGi Core. This dependency should not be used by users. -->
             <dependency>
                 <groupId>org.osgi</groupId>
                 <artifactId>osgi.core</artifactId>
                 <version>${osgi.version}</version>
+                <scope>provided</scope>
+            </dependency>
+
+            <!-- Modular equivalent of osgi.annotation, i.e. the annotation processing of OSGi Core. -->
+            <dependency>
+                <groupId>org.osgi</groupId>
+                <artifactId>org.osgi.annotation.bundle</artifactId>
+                <version>${org.osgi.annotation.bundle.version}</version>
+                <scope>provided</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.osgi</groupId>
+                <artifactId>org.osgi.annotation.versioning</artifactId>
+                <version>${org.osgi.annotation.versioning.version}</version>
+                <scope>provided</scope>
+            </dependency>
+
+            <!-- Modular equivalent of osgi.core, i.e. the run-time of OSGi Core. -->
+            <dependency>
+                <groupId>org.osgi</groupId>
+                <artifactId>org.osgi.dto</artifactId>
+                <version>${org.osgi.dto.version}</version>
+                <scope>provided</scope>
+            </dependency>
+            <dependency>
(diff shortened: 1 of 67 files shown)


config/command/pom.xml
```diff
@@ -53,18 +53,15 @@
     <dependencies>
         <dependency>
             <groupId>org.osgi</groupId>
-            <artifactId>osgi.core</artifactId>
-            <scope>provided</scope>
```
mattrpav, line 57: Why remove 'provided' scope?

bom/pom.xml
```diff
@@ -97,8 +190,8 @@
             </dependency>
             <dependency>
                 <groupId>org.osgi</groupId>
-                <artifactId>org.osgi.namespace.service</artifactId>
```
jbonofre, line 100: I guess we still need `namespace`.

bundle/core/pom.xml
```diff
@@ -56,16 +56,22 @@
...
             <scope>provided</scope>
         </dependency>
-		
-		<dependency>
+
+        <dependency>
             <groupId>org.osgi</groupId>
-            <artifactId>osgi.core</artifactId>
-            <scope>provided</scope>
+            <artifactId>org.osgi.framework</artifactId>
```
jbonofre, line 62: The scope should be `provided` here.

bundle/core/pom.xml
```diff
@@ -56,16 +56,22 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.osgi</groupId>
+            <artifactId>org.osgi.resource</artifactId>
```
jbonofre, line 66: The scope should be `provided` here.

bundle/springstate/pom.xml
```diff
@@ -59,8 +59,7 @@
 
 		<dependency>
             <groupId>org.osgi</groupId>
-            <artifactId>osgi.core</artifactId>
-            <scope>provided</scope>
+            <artifactId>org.osgi.framework</artifactId>
```
jbonofre, line 62: The scope should be `provided` here.

config/command/pom.xml
```diff
@@ -53,18 +53,15 @@
...
+            <artifactId>org.osgi.framework</artifactId>
```
jbonofre, line 56: The scope should be `provided` here.

config/command/pom.xml
```diff
@@ -53,18 +53,15 @@
...
         </dependency>
         <dependency>
             <groupId>org.osgi</groupId>
             <artifactId>org.osgi.service.cm</artifactId>
```
jbonofre, line 60: The scope should be `provided` here.

jbonofre (commented):
Why changing the scope ? Any issue/context for that ?

config/command/pom.xml
rovarga: The declaration in bom specifies scope=provided, so all users get that for free.

jbonofre: @rovarga ok, the bom already defines `provided` by default. Got it. Thanks.

bundle/core/pom.xml
jbonofre: I checked and it's OK here.

jbonofre (approved):

PR #1938, opened 2025-04-24
KARAF-7951: modularize karaf-bom

rovarga:
Split karaf-bom into multiple parts, so users can pick their buy-in level.


jbonofre:
There's something wrong in version resolution with JDK20. I gonna try locally.

mattrpav:
This probably should be Karaf 4.5.x or higher. Adding to 4.4.x would break existing users dependency management and packaging.

jbonofre:
Yes definitely not for 4.4.x. Also I would like to evaluate that.

PR #1942, opened 2025-04-26
[KARAF-7789] Bump Apache POM to 34

rovarga:
The bump to version 32 implies that maven.compiler.release
is also set, which means stricter rules for builds with Java 9+.

We do this in piece multiple parts, cleaning up obvious cases first:
- **Do not use RegistryImpl**
- **Un-modernize GraphQLWebSocketExample**

Next we improve our control over build/runtime enviroment versions:
- **Control m-c-p/m-j-p through property**
- **Add karaf.runtime.target**

Finally leap into the 'building post-Java 9' world:
- **Require Java 11+ for management.server**
- **Require Java 11+ for diagnostic.boot**
- **[KARAF-7789] Bump Apache POM to 34**

### DIFF
--- a/archetypes/assembly/src/main/resources/archetype-resources/pom.xml
+++ b/archetypes/assembly/src/main/resources/archetype-resources/pom.xml
@@ -174,7 +174,7 @@
                         <!-- minimal distribution -->
                         <!--<feature>minimal</feature>-->
                     </bootFeatures>
-                    <javase>1.8</javase>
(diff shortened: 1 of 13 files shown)


jbonofre:
I think the management server needs another change. I will take a look.

jbonofre:
I close this one in favor of a new one I'm working on upgrading to Apache POM 35.

PR #1943, opened 2025-04-26
[KARAF-7767] Set maven.compiler.target=11

rovarga:
Switch to requiring Java 11 by default. Remaining overrides:
- main/pom.xml due to old PAX TinyBundles
- specs/{locator,java.xml,java.xml.ws}/pom.xml for compatibility


jbonofre:
I don't think it's fully correct, and it's breaking the JMX security guard. I will share the needed changes with you.

jbonofre:
Superseeded by #1964

PR #1944, opened 2025-05-07
[KARAF-7856] Improve support for JAAS roles that extend Karaf's RolePrincipal

mattrpav:


shell/ssh/src/test/java/org/apache/karaf/shell/ssh/ExtendedRolePrincipal.java
```diff
@@ -0,0 +1,13 @@
+package org.apache.karaf.shell.ssh;
```
jbonofre, line 1: ASF header is missing here.

mattrpav: fixed in latest push (just now)

jbonofre (approved):

jbonofre (approved):

PR #1945, opened 2025-05-07
[KARAF-7671] Fix LDAP cache handling so it does not reset on every login attempt

mattrpav:
- Move initial cache clear to the Activator
 - Add metrics to the LDAPCache object
 - Fix unit test to _not_ disable cache
 - Update unit test to verify metrics counts and caching are working

NOTE: Caches may pile up when repeatedly deployed via blueprint (since no Activator). This shouldn't be an issue, since even 100 deploys would be small memory impact, and eventually all the caches get cleared.

### DIFF
--- a/jaas/modules/src/main/java/org/apache/karaf/jaas/modules/impl/Activator.java
+++ b/jaas/modules/src/main/java/org/apache/karaf/jaas/modules/impl/Activator.java
@@ -56,6 +56,7 @@ public class Activator extends BaseActivator implements ManagedService {
     @Override
     protected void doOpen() throws Exception {
         super.doOpen();
+        LDAPCache.clear();
         register(BackingEngineFactory.class, new PropertiesBackingEngineFactory());
         register(BackingEngineFactory.class, new PublickeyBackingEngineFactory());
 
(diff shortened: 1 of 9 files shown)


mattrpav:
@jbonofre this change wasn't merged into v4.4.8. Was that intentional?

jbonofre:
No it should have been. I will include in a 4.4.9. Probably the corresponding jira was not targeted to 4.4.8. We have to set the target versions to be sure to include it.

mattrpav:
@jbonofre would you please go ahead and merge to main and 4.4.x branch once the tests pass? This will allow referencing the -SNAPSHOT builds for further testing.

jbonofre:
Sure thing. Give a bit of time as I'm on vacation right now :)

globalbus:
@mattrpav 
I also modified LDAPCache behaviour on my own.
I'm only clearing cache pool on Activator stop, not on start.

The problem is to evict cache. When I used ApacheDS, it support event protocol to launch ObjectChangeListener (that has own drawbacks, like it's need a connection without read timeout). OpenLDAP does not provide support for this.
I finished with custom webhook to clear cache (propagated by JMS to all nodes). But cache can be cleared for all users, not only for changed one, so it's suboptimal.

Anyway, LDAPCache "static" field with caches is a pain in ass. Only way to extend functionality is to copy/paste/fork code. But I don't have better idea for it at the moment. Maybe we could put an instance to javax.security.auth.login.AppConfigurationEntry.options and fetch it from there, instead of static methods? I'm not tried this approach yet.

mattrpav:
@globalbus the cache needs to be cleared on any config change, data change and connection status change. If the LDAP server changes (connect-disconnect), it is not safe to presume the data will be the same.

I'm not married to the static class, but injecting anything into Karaf's JAAS module system is complicated to get consistent via all the provisioning methods: cfg file, blueprint deployment or scr deployment.

I'm leaning to keep this change to be "fix" what is broken and then discuss a refactor/improvement as a follow-on activity.

jbonofre:
I took the liberty to rebase this PR to trigger GitHub Action CI 😄

PR #1964, opened 2025-07-13
KARAF-7889: Upgrade to Apache POM 35 and target Java SE 11

jbonofre:
I have to refactor the JMX connector to work with any Java version.

### DIFF
--- a/specs/java.xml.ws/pom.xml
+++ b/specs/java.xml.ws/pom.xml
@@ -30,6 +30,10 @@
     <artifactId>org.apache.karaf.specs.java.xml.ws</artifactId>
     <name>Apache Karaf :: Specs :: Java Xml WS</name>
 
+    <properties>
+        <maven.compiler.target>8</maven.compiler.target>
+    </properties>
+
     <dependencyManagement>
         <dependencies>
             <dependency>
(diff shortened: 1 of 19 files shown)


specs/java.xml.ws/pom.xml
```diff
@@ -30,6 +30,10 @@
     <artifactId>org.apache.karaf.specs.java.xml.ws</artifactId>
     <name>Apache Karaf :: Specs :: Java Xml WS</name>
 
+    <properties>
+        <maven.compiler.target>8</maven.compiler.target>
```
skitt, line 34: Why is this (and the following) targeting 8 rather than 11?

jbonofre: Some specs still need Java 8 but specs gonna removed anyway.

jbonofre:
I'm fixing the Maven resolver issue.

jbonofre:
I have to ignore the MavenResolver itests for now as they need to update tinybundles version (to 4.x) in order to support Java SE 11+.
However, the CI is OK now. I propose to merge that and work on tinybundles update in a separate PR (I have already a PR about Tinybundles).

PR #1965, opened 2025-07-24
Prepare GH Actions use

jbonofre:
This PR also anticipates a bit the next step by enabling GH Issues.


mattrpav:
@jbonofre sounds good!

PR #1984, opened 2025-08-14
KARAF-7999: expose jdk.jfr packages

rovarga:
Expose the two packages from jdk.jfr module.


jbonofre:
@rovarga thanks ! much appreciated 😄

PR #1985, opened 2025-08-14
Update jre.properties correctly export packages

rovarga:
This is a number of commits, correcting packages to match
their production introduction in Java versions 11-25.

This reverts a number of previous commits, especially around
java.net.http and java.lang.runtime: these were incorrect bug
reports caused by wrong javase (which still defaults to 8!)
requirement, as noted in
https://issues.apache.org/jira/browse/FELIX-6538.

- **KARAF-8000: use cascading properties**
- **KARAF-8000: expose jdk.nio on Java 11+**
- **KARAF-8000: correct Java 12 packages**
- **KARAF-8000: expose jdk.nio.mapmode on Java 14+**
- **KARAF-8000: correct Java 24 packages**
- **KARAF-8000: expose javax.sound on Java 25+**
- **KARAF-8000: correct java.net.http availability**
- **KARAF-8000: correct java.lang.runtime availability**


mattrpav:
This looks like really good coverage!  Good to see the packages in between the incremental version bumps.

jbonofre (approved):
LGTM, thanks !

Can you please rebase (in order to trigger GitHub Action CI) ?

PR #2070, opened 2025-09-29
KARAF-8001: catch UnsupportedOperationException in LogService as it can happen in the case of static distribution

jbonofre:


jbonofre:
@rovarga thanks, I'm creating the PR for the `karaf-4.4.x` branch.

PR #2073, opened 2025-09-29
Fix incomplete upgrade of commons-lang3

holgerfriedrich:
This is required to get the use-base-version test in karaf-maven-plugin pass.
I don't know why the CI passed at that time.

This is the error the PR is fixing:
```
[INFO]   junit.framework.AssertionFailedError: org.custommonkey.xmlunit.Diff
[different] Expected text value 'mvn:org.apache.commons/commons-lang3/3.4' but was 'mvn:org.apache.commons/commons-lang3/3.18.0' - comparing <bundle ...>mvn:org.apache.commons/commons-lang3/3.4</bundle> at /features[1]/feature[1]/bundle[1]/text()[1] to <bundle ...>mvn:org.apache.commons/commons-lang3/3.18.0</bundle> at /features[1]/feature[1]/bundle[1]/text()[1]
```

To see that this is the root cause, one needs to remove the try-catch blocks from the tests.
Then the tests print the exception when failing.
I will create another PR for this.


jbonofre:
Good catch. We actually have a problem on GitHub Action: it executes the tests, but not the itests. I'm fixing that in a separate PR.

PR #2155, opened 2025-11-18
Upgrade to JLine 3.22.0

gnodet:


jbonofre:
@gnodet I confirm that it works. The TCCL the `JansiTerminalProvider` is correctly loaded and the completion works fine.
I propose to merge this PR, I will do the update to the latest JLine version in another PR.

Thanks !

gnodet:
The hack about setting the TCCL to `TerminalBuilder.class.getClassLoader()` should not really be needed, this needs to be improved in JLine.

Issue #2082, opened 2025-10-01
Rework JavaSecurityTest


holgerfriedrich:
`itests/test/src/test/java/org/apache/karaf/itests/JavaSecurityTest.java` has been disabled during the upgrade to JDK21.
See discussion https://github.com/apache/karaf/pull/1992#issuecomment-3353962567.

We have some issues with the `SecurityManager`, which has been deprecated for removal in JDK 17, disabled by default in JDK 21, and permanently disabled in JDK 25 (I refer to the LTS versions only).

This ticket about investigating how Apache Felix handles the removal of SecurityManager in recent JDKs and to derive what is needed on Karaf side.

mattrpav:
Related FELIX ticket: https://issues.apache.org/jira/browse/FELIX-6807
Related FELIX JDK 25 discussion: https://github.com/apache/felix-dev/pull/433

PR #2214, opened 2025-12-30
Lift minimum JDK version to 21 and remove SecurityManager

holgerfriedrich:
This is a first step to prepare the 4.5 branch for running with Java 25.
Starting Karaf in Java 25 does not yet work, as a few functions related to the SecurityManager have been deprecated long time ago and cannot be activated any longer via command line switches.
As a new API has been added in Java 18, I for new lift the minimum required version to Java 21 (LTS).

Maybe this is not the way to go - as there has been some discussions if we could introduce an abstraction layer to allow Java 17 as well.

* Lift minimum JDK version to 21
* Remove occurrences of SecurityManager, as getSecurityManager returns
  null since JDK 18
* ~~Disable JavaSecurityTest, #2082~~
* Enable verbose deprecation warnings
* Replace getSubject and doAs calls
  ** getSubject can no longer be activated via JRE command line as
      SecurityManager has been removed in JDK 25.
      A replacement is available since Java 21.
  ** doAs was deprecated and is replaced by callAll.

For Java 25, another set of modifications is still needed and will be submitted as another PR.
Locally, I got Karaf compiling, passing the tests, and starting with Java 25. 🥳


assemblies/apache-karaf-integration/pom.xml
```diff
@@ -187,7 +187,7 @@
                     </bootFeatures>
                     <libraries>
                     </libraries>
-                    <javase>17</javase>
+                    <javase>21</javase>
```
jbonofre: That's require change on the karaf-maven-plugin and dependencies. I will create another PR (as planned) to tackle that.

itests/test/src/test/java/org/apache/karaf/itests/JavaSecurityTest.java
```diff
@@ -39,6 +40,7 @@
 import static org.ops4j.pax.exam.CoreOptions.maven;
 import static org.ops4j.pax.exam.karaf.options.KarafDistributionOption.*;
 
+@Ignore("Disabled due to removal of SecurityManager. ref Apache Karaf GitHub issue #2082")
```
jbonofre: I would rather remove the test (to avoid confusion) or refactor it instead of ignoring it.

holgerfriedrich:
@jbonofre is there anying else open I should address?

jbonofre:
@holgerfriedrich I think this PR is pretty good, it just needs dependent PRs. I will tackle that.

jbonofre:
@holgerfriedrich do you mind to share your email address with me ? I just realize that you didn't subscribe on the dev mailing list, and you are not on The ASF slack Karaf channel. Thanks !

arusevm:
Hi @holgerfriedrich @jbonofre,

I've got here some ugly patches that seem to make the Subject-related stuff work on java 25.

Basically, there's an issue with that in 25, Subject.callAs() and current() use a ScopedValue and its binding "gets lost" across "thread-boundaries", of which there are a bunch :) It gets ugly, because every time you do something like ExecutorService.submit() you have to also "pass" the current Subject. Gets even worse, because Gogo needs a similar treatment...

Anyway, wondering if it might help if I share what we've got here with you... or if you already have something like it? Or if I might help in some other way? :)

jbonofre:
@arusevm I'm finalizing camel-karaf release. I will be back on Karaf just after, tackling this.

Sorry for the delay.

jbonofre:
Thanks @holgerfriedrich for the update ! I will do a new pass because I have a bunch of changes on the way.

holgerfriedrich:
@jbonofre may I ask you to have a look at this again?

I rebased it in the last days and got the CI green.

It is a major step towards Java 25 compatibility.
Compilation with JDK 25 already works; one test needs a class downgrade and a few itests fail.
I'd like to keep that in a separate PR on top of this one.

~~~
❯ ./bin/karaf
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::staticFieldOffset has been called by org.apache.felix.framework.util.SecureAction (file:/C:/Users/holgerf/src/karaf/assemblies/apache-karaf/target/assembly/system/org/apache/felix/org.apache.felix.framework/7.0.5/org.apache.felix.framework-7.0.5.jar)
WARNING: Please consider reporting this to the maintainers of class org.apache.felix.framework.util.SecureAction
WARNING: sun.misc.Unsafe::staticFieldOffset will be removed in a future release
        __ __                  ____
       / //_/____ __________ _/ __/
      / ,<  / __ `/ ___/ __ `/ /_
     / /| |/ /_/ / /  / /_/ / __/
    /_/ |_|\__,_/_/   \__,_/_/

  Apache Karaf (4.5.0-SNAPSHOT)

Hit '<tab>' for a list of available commands
and '[cmd] --help' for help on a specific command.
Hit '<ctrl-d>' or type 'system:shutdown' or 'logout' to shutdown Karaf.

@root()> shell:info
shell:info
Karaf
  Karaf version               4.5.0-SNAPSHOT
  Karaf home                  C:\Users\holgerf\src\karaf\assemblies\apache-karaf\target\assembly
  Karaf base                  C:\Users\holgerf\src\karaf\assemblies\apache-karaf\target\assembly
  OSGi Framework              org.apache.felix.framework-7.0.5

JVM
  Java Virtual Machine        OpenJDK 64-Bit Server VM version 25+36-3489
  Version                     25
~~~

jbonofre:
@holgerfriedrich I removed the specs bundles. Can you please rebase?

What's your take about JDK 17 / JDK 21?
Personaly, I think it's completely ok to bump directly to JDK 21 for Karaf 4.5.x (JDK 17 is supported at runtime by Karaf 4.4.x).

holgerfriedrich:
@jbonofre if we do not consider writing a compat layer on our own, we need then new API available from JDK18. So 21 LTS is the natural choice.
I have rebased now (mainly due to the deleted files in the spec folder), but did not have time to give it a try in JDK25.

jbonofre:
@holgerfriedrich I agree. I'm advocating for JDK 21 for Karaf 4.5.x. I propose to go that way (including updating our CI).

holgerfriedrich:
@jbonofre how shall we proceed with this one?
Is something missing, or  do you have another approach in mind?

kamilkrzywanski:
@holgerfriedrich there is needed wersion of aries proxy compatible with jdk 25, it seems that their repo is already compatible with jdk 25 but no new release yet.
            <dependency>
                <groupId>org.apache.aries.proxy</groupId>
                <artifactId>org.apache.aries.proxy</artifactId>
                <version>1.1.14</version>
            </dependency>

jbonofre:
@holgerfriedrich @kamilkrzywanski I think we can move forward with this PR and tackle more required updates in follow up PRs (I'm working on Aries releases).

CptBartender:
I raised a PR to @holgerfriedrich's branch with some minor changes that should make it easier to bump Java versions in the future - please consider including it here as well: https://github.com/holgerfriedrich/karaf/pull/176

holgerfriedrich:
Thanks, I will have a look soon.

holgerfriedrich:
@jbonofre I think we are ready. Karaf does not yet start up running with  Java 25, likely due to aries jmx or proxy; I will try to put the current repo build and check. With Java 21, it seems to work.

holgerfriedrich:
I have now compiled aries (quite a few changes necessary due to java 6 to be lifted to java 8) and switched proxy and jmx to snapshot version.
This did not resolve the biggest remaining problem (which occurs once I run Karaf with Java25, it works with Java 21):

grafik

It currently returns an empty user name instead of Karaf, and `bundle:list -t 0` does not show more packages. `bundle:info` with smaller numbers but does give an error in the log like this:
~~~
java.lang.IllegalArgumentException: No matching bundles
        at org.apache.karaf.bundle.command.BundlesCommand.doExecute(BundlesCommand.java:60) ~[?:?]
        at org.apache.karaf.bundle.command.BundlesCommand.execute(BundlesCommand.java:55) ~[?:?]
        at org.apache.karaf.shell.impl.action.command.ActionCommand.execute(ActionCommand.java:88) ~[?:?]
        at org.apache.karaf.shell.impl.console.osgi.secured.SecuredCommand.execute(SecuredCommand.java:68) ~[?:?]
        at org.apache.karaf.shell.impl.console.osgi.secured.SecuredCommand.execute(SecuredCommand.java:86) ~[?:?]
~~~

@jbonofre I think I need some help to make this work....

holgerfriedrich:
Progress: username could be fixed, bundle list still broken... WIP

holgerfriedrich:
Ok, I got some help form AI. Replacing the methods is not enough, we need to adapt the startup of the threads as well.
With the last commit, everything seems to work fine! 🥳 
I removed the patched aries versions, as I could not find any difference in my (small scoped) tests.

I checked the karaf distro both on Java 21 and Java 25. No errors in the log, `bundle:list -t 0` works as expeced.

jbonofre:
@holgerfriedrich I will be back on this one with you to finalize it.

scootermorris:
Any progress on this?  The Cytoscape project is trying to upgrade to Java 25 and right now, this PR seems to be what we need to start moving in that direction.   Once this is committed, we should be able to use the nightly builds to start our testing and (if necessary) make the required changes to our REST infrastructure.

jbonofre:
@scootermorris I'm back on this one for Karaf 4.5.0. I will push some improvements/fixes and we will good to go. Let's say mid next week.

scootermorris:
@jbonofre Great to hear!  Looking forward to it.

holgerfriedrich:
@jbonofre Should I rebase again to check?

holgerfriedrich:
@jbonofre I had to rework, since the files added in the meantime need changes as well.
Adding j25/ubuntu preview/mvn4 build, I stumbled upon an incompatibility: aries proxy package is to java21 byte code. A possible fix seems already been added, but never published:  
https://github.com/apache/aries/blame/6f3c176af5449c7ff38445c5b4c5a16a0bdbf27a/proxy/proxy-impl/src/main/java/org/apache/aries/proxy/impl/ProxyUtils.java#L42
Anything we could do about that?

jbonofre:
@holgerfriedrich thanks for that! I'm working with Domonik for a new Aries Proxy release including the change.

fabgio:
@jbonofre  any news about Aries update? 
@holgerfriedrich  OpenHab contributor here!
Thank you both!

jbonofre:
@fabgio I'm on Aries stuff and related. I will get back to that later today or tomorrow: I'm busy with Karaf 4.4.12 release prep right now.

jbonofre:
@fabgio Karaf 4.5.0 (including this PR) will be my top priority as soon as Karaf 4.4.12 is out.

holgerfriedrich:
I did some testing in the past with the (already committed but not yet published) `aries.proxy-1.1.15-SNAPSHOT`. Apache aries needs to be cloned, then proxy installed. I did this with Java 21:
```
mvn install -N -f parent/pom.xml
mvn install -Daries.skip.version.check=true -f proxy/pom.xml
```

From here you can continue with Java 25 default:

Then it needs a small patch to bring in the new version and additional dependencies:
~~~
diff --git a/assemblies/features/standard/src/main/feature/feature.xml b/assemblies/features/standard/src/main/feature/feature.xml
index 06bb632a51..62a86fa849 100644
--- a/assemblies/features/standard/src/main/feature/feature.xml
+++ b/assemblies/features/standard/src/main/feature/feature.xml
@@ -37,6 +37,8 @@
         <bundle dependency="false" start-level="20">mvn:org.ow2.asm/asm-tree/${asm.version}</bundle>
         <bundle dependency="false" start-level="20">mvn:org.ow2.asm/asm-analysis/${asm.version}</bundle>
         <bundle dependency="false" start-level="20">mvn:org.ow2.asm/asm-commons/${asm.version}</bundle>
+        <!-- aries proxy >= 1.1.15 needs objenesis although its import is optional -->
+        <bundle dependency="false" start-level="20">mvn:org.objenesis/objenesis/${objenesis.version}</bundle>
         <bundle start-level="20">mvn:org.apache.aries.proxy/org.apache.aries.proxy/${aries.proxy.version}</bundle>
     </feature>
 
diff --git a/bom/pom.xml b/bom/pom.xml
index 225a9ed2e1..38ea526eac 100644
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -478,6 +478,11 @@
                 <artifactId>asm-commons</artifactId>
                 <version>${asm.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.objenesis</groupId>
+                <artifactId>objenesis</artifactId>
+                <version>${objenesis.version}</version>
+            </dependency>
 
             <dependency>
                 <groupId>org.apache.felix</groupId>
diff --git a/itests/test/src/test/filtered-resources/etc/feature.xml b/itests/test/src/test/filtered-resources/etc/feature.xml
index b4c139af3d..7a0c9d69ab 100644
--- a/itests/test/src/test/filtered-resources/etc/feature.xml
+++ b/itests/test/src/test/filtered-resources/etc/feature.xml
@@ -37,6 +37,7 @@
         <bundle dependency="false" start-level="20">mvn:org.ow2.asm/asm-tree/${asm.version}</bundle>
         <bundle dependency="false" start-level="20">mvn:org.ow2.asm/asm-analysis/${asm.version}</bundle>
         <bundle dependency="false" start-level="20">mvn:org.ow2.asm/asm-commons/${asm.version}</bundle>
+        <bundle dependency="false" start-level="20">mvn:org.objenesis/objenesis/${objenesis.version}</bundle>
         <bundle start-level="20">mvn:org.apache.aries.proxy/org.apache.aries.proxy/${aries.proxy.version}</bundle>
     </feature>
 
diff --git a/pom.xml b/pom.xml
index 763c497567..ece6f5933d 100644
--- a/pom.xml
+++ b/pom.xml
@@ -263,7 +263,7 @@
         <aries.jndi.api.version>1.1.0</aries.jndi.api.version>
         <aries.jndi.core.version>1.0.2</aries.jndi.core.version>
         <aries.jndi.url.version>1.1.0</aries.jndi.url.version>
-        <aries.proxy.version>1.1.14</aries.proxy.version>
+        <aries.proxy.version>1.1.15-SNAPSHOT</aries.proxy.version>
         <aries.proxy.api.version>1.1.1</aries.proxy.api.version>
         <aries.spifly.version>1.3.7</aries.spifly.version>
         <aries.subsystem.api.version>2.0.10</aries.subsystem.api.version>
@@ -292,6 +292,7 @@
         <maven.wagon.version>3.5.3</maven.wagon.version>
         <maven-plugin-annotations.version>3.16.0</maven-plugin-annotations.version>
         <maven.resolver.version>2.0.23</maven.resolver.version>
+        <objenesis.version>3.6</objenesis.version>
 
         <osgi.version>8.0.0</osgi.version>
         <org.osgi.annotation.bundle.version>2.0.0</org.osgi.annotation.bundle.version>
~~~

This makes all our tests work on a Java 25.
Note that we still lock the code generation to Java 21 in Karaf.

For newer Java versions, aries needs another patch, apache/aries#912. I tried this and got it working with Java 27.
It matters on which JDK aries proxy runs - not the class file version which is generated.

scootermorris:
Hi all.  Are we getting any closer to a 4.5-SNAPSHOT that supports Jdk 25?  The natives are getting restless out in Cytoscape-land...

jbonofre:
@scootermorris I share again: the focus is on Karaf 4.4.12 right now, 4.5.0 (with JDK support update) will follow.

PR #1960, opened 2025-07-06
KARAF-7974: Upgrade to Pax URL 3.0.1

jbonofre:

PR #2260, opened 2026-02-04
[karaf-4.4.x] KARAF-7974: Upgrade to Pax URL 3.0.1 (#1960)

jbonofre:


jbonofre:
@grgrzybek the change has been made on `main` with updated handler.

grgrzybek:
may I see the change?

jbonofre:
The problem is that how to deal with both `javax.servlet` vs `jakarta.servlet`.

As Karaf (including `main`) is still Pax Web 8, it means `javax.servlet` still. So, my change was just to "downgrade" the namespace.

PR #2322, opened 2026-03-14
Prepare for Maven 4

holgerfriedrich:
This basically allows building Karaf with maven4-rc5.

First addition is the directory `.mvn` to mark the top directory of the workspace. Not strictly necessary, but it suppresses the warning about missing top level directory marker during build with Maven 4.
The approach to add .mvn dir is the option that does not impact our standard builds with Maven 3.

The second addition is a missing version specification in a pom file. Without it, Maven4 just picks the wrong version of the maven-resources-plugin, which is built against an old API and is causing the build to fail.

We still get warnings about BOM import. This is a bigger change an currently does not break the build. It is not in scope of this PR.

### DIFF
--- a/archetypes/kar/src/main/resources/archetype-resources/pom.xml
+++ b/archetypes/kar/src/main/resources/archetype-resources/pom.xml
@@ -69,6 +69,11 @@
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-resources-plugin</artifactId>
+                <version>3.3.1</version>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>3.5.2</version>
             </plugin>
--- a/.mvn/README
+++ b/.mvn/README
@@ -1,0 +1,1 @@
+required by Maven 4 to detect root dir; adding root=true to pom is not compatible with Maven 3


jbonofre (approved):

holgerfriedrich:
@jbonofre Can you please advise who should merge? The author once a PR is approved?
(Ok in thins case, it first needs a rebase due to the new branch protection rules....)

jbonofre:
@holgerfriedrich As soon as we have one approval, the author or the reviewer can merge.

archetypes/kar/src/main/resources/archetype-resources/pom.xml
```diff
@@ -67,6 +67,11 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.13.0</version>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-resources-plugin</artifactId>
+                <version>3.3.1</version>
```
jbonofre, line 73: This comes from Apache POM. Do we really need to override the version here ?

holgerfriedrich: I just gave it a try and is is actually required.

Anyhow, in this POM, the versions for all plugins are set. And the 3.3.1 for the maven-resources-plugin seem to be set in other places througout the codebase as well....

jbonofre: Oh, it's in the archetypes. You are right, that's normal because archetypes don't inherit from Apache POM (and it's by intend).

PR #2327, opened 2026-03-14
Upgrade Maven plugins in archetype pom files

holgerfriedrich:
It seems that dependabot cannot handle updates in those files as they are some kind of template containing variables etc.

For now a draft, not yet tested.
btw: those templates contain also hard coded java target version 1.8 for the maven-compiler-plugin.

@jbonofre FYI

### DIFF
--- a/archetypes/kar/src/main/resources/archetype-resources/pom.xml
+++ b/archetypes/kar/src/main/resources/archetype-resources/pom.xml
@@ -70,7 +70,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-resources-plugin</artifactId>
-                <version>3.3.1</version>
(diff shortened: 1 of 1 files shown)


jbonofre:
That looks good to me. Let's keep the archetypes for Karaf 4.x, we will revisit if it still makes sense for Karaf 5.x.

Thanks!

holgerfriedrich:
Ok thanks. I have checked the codebase for more occurrences. It seems fine as it is.

holgerfriedrich:
Either this PR or #2322 will need a rebase. Just let me know with which one you want to proceed first.

holgerfriedrich:
Rebased, added one missing occurrence, finally then CI is green..

Shall we merge?

jbonofre:
👍

PR #2516, opened 2026-04-05
Remove static ThreadLocal caching in XmlUtils to prevent ClassLoader leaks

### LINKED ISSUE #2278: ClassLoader leak and "Zombie Bundles" caused by static ThreadLocals in XmlUtils
Describe the bug 
I identified a structural defect (UTL-14) in org.apache.karaf.util.XmlUtils. The class uses static ThreadLocal variables to cache XML factory instances (DocumentBuilderFactory, TransformerFactory, SAXParserFactory) without any cleanup mechanism (remove() is never called).

In an OSGi environment (Karaf), these factory instances often originate from specific bundles (e.g., Xerces or system bundle fragments). Holding them in static ThreadLocals of long-lived Karaf worker threads pins the ClassLoader of the provider bundle.

Source Code Location:
// org/apache/karaf/util/XmlUtils.java (Lines 51-53)
private static final ThreadLocal<DocumentBuilderFactory> DOCUMENT_BUILDER_FACTORY = new ThreadLocal<>();
private static final ThreadLocal<TransformerFactory> TRANSFORMER_FACTORY = new ThreadLocal<>();
private static final ThreadLocal<SAXParserFactory> SAX_PARSER_FACTORY = new ThreadLocal<>();

// Usage example (Line 132):
// .set() is called, but .remove() is never called anywhere in the class.
TRANSFORMER_FACTORY.set(tf);

Impact

1. OSGi Bundle Locking (Zombie Bundles): If the bundle providing the XML implementation is updated or refreshed, the old ClassLoader cannot be garbage collected because it is strongly referenced by the ThreadLocalMap of active threads.

2. Metaspace Leak: Repeated updates of XML-related bundles will lead to permanent Metaspace growth.

3. Stale Logic: Threads may continue to use XML factories from an uninstalled/obsolete bundle version.

Expected behavior 
XmlUtils should avoid using static ThreadLocal for holding OSGi-dependent resources, or ensure that remove() is called after usage. Given that XmlUtils is a static utility class with no lifecycle management, removing the ThreadLocal caching (instantiating factories per request) or using a SoftReference approach is recommended to ensure OSGi stability.

Actual behavior 
The XML factories are cached indefinitely on threads, preventing the underlying provider bundles from being fully unloaded/refreshed.

Suggested Fix
Remove the static ThreadLocal caching entirely. Although creating XML factories has a small performance cost, correct ClassLoader handling in OSGi is critical.

Proposed Change:
// Remove static ThreadLocals
// private static final ThreadLocal<DocumentBuilderFactory> DOCUMENT_BUILDER_FACTORY = new ThreadLocal<>();
// ...

public static DocumentBuilder documentBuilder() throws ParserConfigurationException {
    // Instead of retrieving from ThreadLocal, create a new instance every time
    // to ensure the correct ClassLoader (from the current context/bundle) is used.
    DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
    dbf.setNamespaceAware(true);
    // ... set other features ...
    return dbf.newDocumentBuilder();
}

Alternatively, if performance is a blocker, use a SoftReference or implement a Closeable hook to clear ThreadLocals on bundle stop, but removing the cache is the safest for static utility classes.

jbonofre:
The static `ThreadLocal` fields holding `DocumentBuilderFactory`, `TransformerFactory` and `SAXParserFactory` pin the ClassLoader of the XML provider bundle on long-lived Karaf worker threads. When that bundle is updated or refreshed, the old ClassLoader can never be GC'd, causing zombie bundles and Metaspace growth.

This patch removes the `ThreadLocal` caching and creates fresh factory instances per call. The public API is unchanged.

Fixes #2278

### DIFF
--- a/util/src/main/java/org/apache/karaf/util/XmlUtils.java
+++ b/util/src/main/java/org/apache/karaf/util/XmlUtils.java
@@ -46,10 +46,6 @@ public class XmlUtils {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(XmlUtils.class);
 
-    private static final ThreadLocal<DocumentBuilderFactory> DOCUMENT_BUILDER_FACTORY = new ThreadLocal<>();
-    private static final ThreadLocal<TransformerFactory> TRANSFORMER_FACTORY = new ThreadLocal<>();
-    private static final ThreadLocal<SAXParserFactory> SAX_PARSER_FACTORY = new ThreadLocal<>();
-
     public static Document parse(String uri) throws TransformerException, IOException, SAXException, ParserConfigurationException {
         DocumentBuilder db = documentBuilder();
         try {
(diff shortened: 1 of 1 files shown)


holgerfriedrich (approved):
LGTM

PR #2531, opened 2026-04-12
fix(ci): Add GH run id to artifacts to prevent race condition

holgerfriedrich:


holgerfriedrich:
Let's see if workflow `ci-test-results.yml` picks up the renamed event file as well....

holgerfriedrich:
This should not be necessary, since this is upload-artifact and not cache 😎 
Sorry for the noise.

PR #2550, opened 2026-04-19
Disable flaky SSH tests on Windows

holgerfriedrich:
This addresses the repeated CI failures after 1h testing on the Windows platfrom.

It is _not fixing_ it, but disables ~3~ 6 SSH tests on Windows.
It could be a viable way to get rid of the failed builds and an alterative to #2544.

@jbonofre Are you thinking of a different approach to the issue, or shall we got in this direction for now?


holgerfriedrich:
Retriggering, in case the green build was just good luck.

jbonofre (approved):
Let's skip the test on Windows for now, I will work on a fix for PTY (probably using the exec channel).

PR #2577, opened 2026-04-24
fix(karaf-maven-plugin): handle KAR files in features-generate-descriptor (4.4.x)

jbonofre:
Commit 9afa6485c5 removed the FEATURE_CLASSIFIER guard in processFeatureArtifact, causing .kar artifacts to enter the feature processing branch. Since a KAR is a ZIP archive, JaxbUtil.unmarshal() failed with "Content is not allowed in prolog" when trying to parse it as raw XML.

Fix readFeaturesFile() to detect .kar files and extract their embedded features XML entries from the repository/ directory inside the archive, merging them into a single Features object.


jbonofre:
@holgerfriedrich that should fix the issue you have with karaf-maven-plugin 4.4.11. That's my bad 🥲 

Do you mind to test that before I merge (and release)?

jbonofre:
The fix has been verified (and approved) by @holgerfriedrich in #2578

PR #2578, opened 2026-04-24
fix(karaf-maven-plugin): handle KAR files in features-generate-descritor

jbonofre:
Commit 9afa6485c5 removed the FEATURE_CLASSIFIER guard in processFeatureArtifact, causing .kar artifacts to enter the feature processing branch. Since a KAR is a ZIP archive, JaxbUtil.unmarshal() failed with "Content is not allowed in prolog" when trying to parse it as raw XML.

Fix readFeaturesFile() to detect .kar files and extract their embedded features XML entries from the repository/ directory inside the archive, merging them into a single Features object.


holgerfriedrich (approved):
tested with my app on top of Karaf: cancelled rc0 fails, with this PR completes the build completes

PR #2579, opened 2026-04-24
Fix inclomplete stax2-api upgrade to 4.3.0

holgerfriedrich:
Refs: #2565


jbonofre (approved):

jbonofre:
So, generally speaking we should not touch the "hard coded" version in features XML, because these versions are really needed to resolve the feature. It doesn't intentionally use Maven property here. I think a better fix would be to use Maven properties where it makes sense (and applicable), but not systemically change the versions in features XML.

holgerfriedrich:
I am fine if we do not proceed with my 2 PRs (feel free to close them).
I just noticed the inconsistency when building my app on top (which resolves automatically to stax2-api 4.2.2 and byte-buddy 1.17.7 due to other dependencies anyway, regardless of this change).

jbonofre:
@holgerfriedrich let me review: if it doesn't break anything, it's worth to update 😄

jbonofre:
Hibernate6 is still happy, and the camel-blueprint is also happy.

However, I'm in favor to revert #2565 just in case on karaf-4.4.x.

PR #2580, opened 2026-04-24
Fix inclomplete byte-buddy upgrade to 1.18.8

holgerfriedrich:
Refs: #2563

### DIFF
--- a/assemblies/features/enterprise/src/main/feature/feature.xml
+++ b/assemblies/features/enterprise/src/main/feature/feature.xml
@@ -200,7 +200,7 @@ com.atomikos.icatch.log_base_dir=${karaf.data}/atomikos
         <feature>aries-blueprint</feature>
         <!-- Core logging and bytecode -->
         <bundle>mvn:org.jboss.logging/jboss-logging/3.5.3.Final</bundle>
-        <bundle>mvn:net.bytebuddy/byte-buddy/1.17.5</bundle>
+        <bundle>mvn:net.bytebuddy/byte-buddy/1.18.8</bundle>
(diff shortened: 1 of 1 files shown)


holgerfriedrich:
@jbonofre there is one more occurrence of and older version of byte-buddy in the same file. It is used in hibernate feature.
Don't know how to proceed with this, it is _very_ old and did not match the 1.17.5 we had before.

jbonofre:
Generally speaking, the "hard coded" version in the features are correct because it matches the expection of the feature. That's the main reason we are using specific version here (not the maven property).
So, I don't think we should update the feature here.

holgerfriedrich:
Ok, lets keep the older one for the hibernate feature.

What about the hibernate6 feature proposed change? 1.17.5 does not match the library requirements in hibernate6 as well (6.6.18-final seems to use 1.15.11). Though, I did not check the other libs in the feature.
1.17.5 -> 1.18.8 is the change which was done in #2563.

jbonofre:
Yes, indeed, I was maybe too agressive in the new hibernate6 feature. It works because the hibernate wrap is using "large range" import but worth to align.
We should check hibernate6 dependencies to align the expected versions.

holgerfriedrich:
I am fine if we do not proceed with my 2 PRs (feel free to close them).
I just noticed the inconsistency when building my app on top (which resolves automatically to stax2-api 4.2.2 and byte-buddy 1.17.7 due to other dependencies anyway, regardless of this change).

jbonofre:
Same as stax2-api, let's revert #2563

PR #2588, opened 2026-04-25
Prepare for Maven 4 (#2322)

holgerfriedrich:
Backport of #2322 

This basically allows building Karaf with maven4-rc5.

First addition is the directory .mvn to mark the top directory of the workspace. Not strictly necessary, but it suppresses the warning about missing top level directory marker during build with Maven 4.
The approach to add .mvn dir is the option that does not impact our standard builds with Maven 3.

The second addition is a missing version specification in a pom file. Without it, Maven4 just picks the wrong version of the maven-resources-plugin, which is built against an old API and is causing the build to fail.

### DIFF
--- a/.mvn/README
+++ b/.mvn/README
@@ -1,0 +1,1 @@
+required by Maven 4 to detect root dir; adding root=true to pom is not compatible with Maven 3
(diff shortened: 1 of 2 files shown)


holgerfriedrich:
@jbonofre a nice to have on 4.4.x, I would appreciate if I could stay at Maven 4 😎 
At least the compilation fail is unfortunate and may affect others.

jbonofre:
@holgerfriedrich it makes sense. As the change is pretty minimal, it's totally fine to have it for 4.4.x.

jbonofre (approved):

PR #2625, opened 2026-04-30
Migrate to Jackson 3.1.2

holgerfriedrich:
* migrate Jackson packages from com.fasterxml.jackson namespace to tools.jackson namespace
* suppress com.fasterxml for hibernate to avoid double inclusion (hibernate should then select tools.jackson packages)
* change of jackson related features

This PR basically tries to remove jackson 2. We still need it for spring.security and two transitive dependencies in examples.


holgerfriedrich:
@jbonofre I think I managed to migrate fully to jackson 3, with no dependencies to jackson 2 visible.
This is not yet pushed to this PR, as it is on top of the pax web and java25 patch.
I'll try to separate out a few more PRs (findings regarding other topics and self-contained preparations for the jackson3 migration).

PR #2627, opened 2026-05-02
More static method conversions

rovarga:
This is another round of marking methods static.

### DIFF
--- a/bundle/core/src/main/java/org/apache/karaf/bundle/core/internal/BundleWatcherImpl.java
+++ b/bundle/core/src/main/java/org/apache/karaf/bundle/core/internal/BundleWatcherImpl.java
@@ -138,12 +138,12 @@ public class BundleWatcherImpl implements Runnable, BundleListener, BundleWatche
         }
     }
 
-    private String getLocation(Bundle bundle) {
+    private static String getLocation(Bundle bundle) {
         String location = bundle.getHeaders().get(Constants.BUNDLE_UPDATELOCATION);
         return location != null ? location : bundle.getLocation();
     }
 
-    private boolean isMavenSnapshotUrl(String url) {
+    private static boolean isMavenSnapshotUrl(String url) {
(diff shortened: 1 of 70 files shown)


bundle/core/src/main/java/org/apache/karaf/bundle/core/internal/BundleWatcherImpl.java
```diff
@@ -138,12 +138,12 @@ public void run() {
         }
     }
 
-    private String getLocation(Bundle bundle) {
+    private static String getLocation(Bundle bundle) {
```
jbonofre, line 141: What is the rationale for all these static?

rovarga: These are pointed out by Eclipse, so it just reduces clutter.

As to the scope/randomness of the addressed methods: I am working on NIO2 conversion patches, so I am visiting a lot of components -- so I thought I'd do these as a drive-by, but keep them separate from other changes to make review easier.

Btw. sorry for not being on Slack, it keeps crashing on me and I did not have the cycles to debug exactly why :(

rovarga: Reviewing the state of that upcoming patch: there are a few other conversions, but strictly in files which are also touched by the NIO2 conversion.

So it's up to you:
- I can update this PR with the follow-up statics
- I can raise a separate PR for just the next round (i.e. delta on top of this PR)
- I can just keep them in the NIO2 patch

jbonofre: Let's update in this PR to have all related changes grouped. Thanks!

rovarga: Ack, incoming.

PR #2649, opened 2026-05-07
Remove use of File{Input,Ouput}Stream

rovarga:
These two classes have a finalizer in Java 12, hence they are an
unnecessary burden for GC.

Use:
- Files.newInputStream() instead of FileInputStream
- Files.newBufferedReader() when we end up converting to a Writer
- Files.readAllBytes() if we end up reading the stream
- Files.copy() if we are copying the stream to another stream or file
- Files.writeString() when we are emitting just a plain string
- Files.newOutputStream() instead of FileOutputStream
- Files.newBufferedReader() when we end up conerting to a Writer

### DIFF
--- a/features/core/src/main/java/org/apache/karaf/features/internal/download/impl/SimpleDownloadTask.java
+++ b/features/core/src/main/java/org/apache/karaf/features/internal/download/impl/SimpleDownloadTask.java
@@ -17,15 +17,11 @@
 package org.apache.karaf.features.internal.download.impl;
 
 import java.io.File;
-import java.io.FileOutputStream;
 import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
 import java.net.URL;
 import java.nio.file.Files;
 import java.util.concurrent.ScheduledExecutorService;
 
-import org.apache.karaf.util.StreamUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -75,9 +71,8 @@ public class SimpleDownloadTask extends AbstractRetryableDownloadTask {
             File tmpFile = Files.createTempFile(dir.toPath(), "download-", null).toFile();
             
             urlObj = new URL(DownloadManagerHelper.stripStartLevel(urlObj.toString()));
-            try (InputStream is = urlObj.openStream();
-                 OutputStream os = new FileOutputStream(tmpFile)) {
-                StreamUtils.copy(is, os);
+            try (var is = urlObj.openStream()) {
+                Files.copy(is, tmpFile.toPath());
             }
 
             if (file.exists() && !file.delete()) {
@@ -112,10 +107,8 @@ public class SimpleDownloadTask extends AbstractRetryableDownloadTask {
         File dir = new File(System.getProperty("karaf.data"), "tmp");
         dir.mkdirs();
         File tmpFile = Files.createTempFile(dir.toPath(), "download-", null).toFile();
-        try (InputStream is = new URL(url).openStream();
-             OutputStream os = new FileOutputStream(tmpFile))
-        {
-            StreamUtils.copy(is, os);
+        try (var is = new URL(url).openStream()) {
+            Files.copy(is, tmpFile.toPath());
         }
         return tmpFile;
     }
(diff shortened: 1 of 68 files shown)


features/core/src/main/java/org/apache/karaf/features/internal/download/impl/SimpleDownloadTask.java
```diff
@@ -75,9 +71,8 @@ protected File download(Exception previousExceptionNotUsed) throws Exception {
             File tmpFile = Files.createTempFile(dir.toPath(), "download-", null).toFile();
             
             urlObj = new URL(DownloadManagerHelper.stripStartLevel(urlObj.toString()));
-            try (InputStream is = urlObj.openStream();
-                 OutputStream os = new FileOutputStream(tmpFile)) {
-                StreamUtils.copy(is, os);
+            try (var is = urlObj.openStream()) {
+                Files.copy(is, tmpFile.toPath());
```
jbonofre, line 75: I think this will always throws `FileAlreadyExistsException` because `Files.creqteTempFile()` creates the file before the copy is attempted. `REPLACE_EXISTING` is required here.

features/core/src/main/java/org/apache/karaf/features/internal/download/impl/SimpleDownloadTask.java
```diff
@@ -112,10 +107,8 @@ protected File downloadBlueprintOrSpring() throws Exception {
         File dir = new File(System.getProperty("karaf.data"), "tmp");
         dir.mkdirs();
         File tmpFile = Files.createTempFile(dir.toPath(), "download-", null).toFile();
-        try (InputStream is = new URL(url).openStream();
-             OutputStream os = new FileOutputStream(tmpFile))
-        {
-            StreamUtils.copy(is, os);
+        try (var is = new URL(url).openStream()) {
+            Files.copy(is, tmpFile.toPath());
```
jbonofre, line 111: I think this will always throws `FileAlreadyExistsException` because `Files.creqteTempFile()` creates the file before the copy is attempted. `REPLACE_EXISTING` is required here.

features/core/src/main/java/org/apache/karaf/features/internal/service/FeatureConfigInstaller.java
```diff
@@ -298,21 +302,15 @@ private void installConfigurationFile(String fileLocation, String finalname, boo
...
                     parentFile.mkdirs();
                 }
                 file.createNewFile();
             }
-            try (
-                    FileOutputStream fop = new FileOutputStream(file)
-            ) {
-                StreamUtils.copy(is, fop);
-            }
+            Files.copy(is, file.toPath());
```
jbonofre, line 313: Without `REPLACE_EXISTING`, it throws when `override=true` and the config file already exists. The old `FileOutputStream` silently overwrote. With `Files.copy()` will always fail the override path. The same happens in `InstallCommand` and `ConfigMBeanImpl`.

kar/src/main/java/org/apache/karaf/kar/internal/Kar.java
```diff
@@ -161,9 +160,7 @@ private static File extract(InputStream is, ZipEntry zipEntry, File dest) throws
             dest.mkdirs();
         } else {
             dest.getParentFile().mkdirs();
-            FileOutputStream out = new FileOutputStream(dest);
-            StreamUtils.copy(is, out);
-            out.close();
+            Files.copy(is, dest.toPath());
```
jbonofre, line 163: I think `REPLACE_EXISTING` is required here. On KAR re-install or upgrade, previously extracted files exist, causing `FileAlreadyExistsException` and leaving the KAR partially extracted.

features/core/src/main/java/org/apache/karaf/features/internal/service/FeatureConfigInstaller.java
```diff
@@ -173,12 +181,8 @@ public void installFeatureConfigs(Feature feature) throws IOException, InvalidSy
...
     }
 
-    private String loadConfiguration(final URL url) throws IOException {
-        try (final InputStream inputStream = new BufferedInputStream(url.openStream())) {
-            final ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
-            StreamUtils.copy(inputStream, outputStream);
-            return new String(outputStream.toByteArray(), StandardCharsets.UTF_8);
-        }
+    private static String loadConfiguration(final URL url) throws IOException {
+        return new String(url.openStream().readAllBytes(), StandardCharsets.UTF_8);
```
jbonofre, line 185: Here, we potentially leaks the URL stream if `readAllBytes()` throws mid-read. Previously, we used a proper try-with-resources approach.

jbonofre (changes requested):

PR #2682, opened 2026-05-27
Added option to configure dynamic imports

CptBartender:
Current `bundle:dynamic-imports` command only allows toggling of dynamic imports. This PR adds an option to configure which packages are imported dynamically, instead of setting all-or-nothing, as well as adds option to explicitly set or unset the option.

### DIFF
--- a/bundle/core/src/main/java/org/apache/karaf/bundle/command/DynamicImport.java
+++ b/bundle/core/src/main/java/org/apache/karaf/bundle/command/DynamicImport.java
@@ -16,26 +16,54 @@
  */
 package org.apache.karaf.bundle.command;
 
+import java.util.List;
+import org.apache.karaf.shell.api.action.Argument;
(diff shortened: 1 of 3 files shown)


jbonofre (approved):

holgerfriedrich:
I'm wondering why this PR was against 4.4.x and not main. Did I overlook something?

jbonofre:
@holgerfriedrich oh right, it should have been on `main` not `karaf-4.4.x`. Let me backport.

CptBartender:
> I'm wondering why this PR was against 4.4.x and not main. Did I overlook something?

This is because I needed that change on 4.4.x and eventually in 4.4.12 release. I don't know the process, and was unsure if I should create a separate PR to `main` or if the entire `karaf-4.4.x` branch will eventually be merged. Apologies for the inconvenience.

PR #2720, opened 2026-06-10
Refactored usage of Optional<> as method parameters

CptBartender:
There's only one place that uses `Optional` as method parameters, and it is generally recommended to be avioded. I've reworked the use case so that the `Optional`s are no longer used as method parameters or generic parameters for other functional interfaces


config/core/src/main/java/org/apache/karaf/config/core/MetaServiceCaller.java
```diff
@@ -73,13 +72,15 @@ public static List<String> getPidsWithMetaInfo(BundleContext context) {
...
      */
-    public static <T> T doWithMetaType(BundleContext context, String pid, Function<Optional<MetaInfo>, T> function) {
+    public static <T> T doWithMetaType(BundleContext context, String pid, Function<MetaInfo, T> function) {
         ServiceReference<MetaTypeService> ref = context.getServiceReference(MetaTypeService.class);
         if (ref != null) {
             try {
                 MetaTypeService metaService = context.getService(ref);
                 MetaInfo metaInfo = getMetatype(context, metaService, pid);
-                return function.apply(Optional.ofNullable(metaInfo));
+                if (metaInfo != null) {
```
holgerfriedrich: This change seems to have an effect on semantics for the null case: before, `function.apply` of an empty `Optional` was returned, now the fall-through may just return null.
This could lead to an NPE in the statement from the other file above (`return MetaServiceCaller.doWithMetaType(...)`).

CptBartender: Good catch - fixed.

holgerfriedrich (approved):
LGTM, thanks.
A new version like 4.5 seems a good point in time to clean up the API,

jbonofre (approved):

PR #2766, opened 2026-07-08
Fix flaky SSH security integration tests on Windows CI (#2764)

jbonofre:
* Fix flaky SSH security itests on Windows by waiting for command completion

The SSH command security integration tests (SshCommandTestBase) wrote a command to the SSH channel and immediately sent "logout" to tear down the session. On slower runners - in particular the Windows CI - the session could be closed while the command output was still in flight, producing truncated output (e.g. a partial command echo "she" instead of "shell:nano") and spurious assertion failures such as "Should contain 'Command not found'".

Introduce a writeCommandAndWait() helper that appends a sentinel "echo <marker>" command and blocks until the unique marker appears in the captured output before closing the channel. Because the remote shell reads and executes its input line by line, the marker cannot appear before the command under test has been fully executed and flushed back to the client, providing a deterministic completion signal that also works for the OK case (which has no positive marker of its own). "echo" is a gogo built-in that is not restricted by any command ACL, so it is safe for every test user.

The helper is also used by addUsers()/addViewer() so the JAAS users are fully created before the test logs in as them.

* Potential fix for pull request finding

---------

### DIFF
--- a/itests/test/src/test/java/org/apache/karaf/itests/ssh/SshCommandTestBase.java
+++ b/itests/test/src/test/java/org/apache/karaf/itests/ssh/SshCommandTestBase.java
@@ -20,6 +20,7 @@ import java.io.PipedInputStream;
 import java.io.PipedOutputStream;
 import java.util.EnumSet;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 
 import org.apache.karaf.itests.BaseTest;
 import org.apache.sshd.client.SshClient;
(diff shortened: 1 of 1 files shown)


holgerfriedrich:
@jbonofre I think we skip the itests on Windows on the 4.4.x branch since #2537, or do I get this wrong?

jbonofre:
@holgerfriedrich #2537 just updated the build contexts (checks) on CI, but it still execute the tests.

PR #2815, opened 2026-08-23
fix(itests): Speed up JVM teardown by using TCP communication instead of stdio

holgerfriedrich:
This tries to mitigate the 30sec timeouts we see in our tests by switching to TCP communcation towards the forked JVMs.
Let's see how much faster it is.

Ref:
https://maven.apache.org/surefire/maven-surefire-plugin/examples/process-communication.html


holgerfriedrich:
This seems to be a good improvement :tada: 

(though only looking at a single sample)
image
vs
image

jbonofre (approved):

PR #2838, opened 2026-08-29
Upgrade bouncy castle / bcprov from 1.85 to 1.85.2

holgerfriedrich:
Patch-level bug-fix release:

https://www.bouncycastle.org/download/bouncy-castle-java/?filter=java%3Drelease-1-85-2


holgerfriedrich:
Only bcprov was upgraded. I do not want to increase complexity by splitting up version numbers for the bc components. Sticking to 1.85 for all components.

PR #2839, opened 2026-08-29
Allow configuring base docker image in DockerfileMojo (#2785)

### LINKED ISSUE #2785: Hardcoded docker image at `DockerfileMojo.java`
The  [`DockerfileMojo`](https://github.com/apache/karaf/blob/d3a178529f965f247d7f9cc0a9cd5af310df5389/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/DockerfileMojo.java#L49) contains the hardcoded docker image `eclipse-temurin:11-jre`.

If we build a karaf assembly with Java 17 and generate the Dockerfile, the assembly will not start (wrong JRE version).

Maybe I'm overlooking something?

Alwaysgaurav1:
Fixes #2785

### Changes:
- Add configurable `@Parameter` `image` (defaulting to `eclipse-temurin:11-jre`) in `DockerfileMojo.java` so assemblies targeting Java 17+ or custom runtimes can specify their base image.
- Add `DockerfileMojoTest.java` unit test suite covering default and custom configurations.
- Strengthen `test-dockerfile` integration test verification in `verify.bsh`.
- Update `karaf-maven-plugin.adoc` documentation to describe `image` and `command` parameters, and fix an XML closing tag typo (`<goal>docker</goal>`).


tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/DockerfileMojo.java
```diff
@@ -40,13 +40,16 @@ public class DockerfileMojo extends MojoSupport {
...
+    private String image;
+
     @Override
     public void execute() throws MojoExecutionException, MojoFailureException {
         getLog().info("Creating Dockerfile");
         File dockerFile = new File(destDir, "Dockerfile");
         try {
             StringBuilder buffer = new StringBuilder();
-            buffer.append("FROM eclipse-temurin:11-jre").append("\n");
+            buffer.append("FROM ").append(image).append("\n");
```
jbonofre: Neither `image` nor `command` are validated before being written into the Dockerfile. An empty `<image></image>` (or `-Dimage=`) would produce an invalid `FROM` line with no argument, and a value containing an embedded newline would silently inject extra Dockerfile instructions.

Worth at least falling back to the default when the value is blank.

Alwaysgaurav1:
Thanks for the feedback, @jbonofre! I have updated the PR to:
1. Fall back to default values (`eclipse-temurin:11-jre` and `["karaf", "run"]`) when `image` or `command` is null or blank.
2. Validate and reject values containing newline characters with a `MojoExecutionException`.
3. Add comprehensive unit tests covering the blank fallbacks and newline validation scenarios.

tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/DockerfileMojo.java
```diff
@@ -31,29 +31,46 @@
...
+    private static final String DEFAULT_COMMAND = "[\"karaf\", \"run\"]";
+
     @Parameter(defaultValue = "${project.build.directory}")
     private File destDir;
 
     @Parameter(defaultValue = "${project.build.directory}/assembly")
     private File assembly;
 
-    @Parameter(defaultValue = "[\"karaf\", \"run\"]")
+    @Parameter(defaultValue = DEFAULT_COMMAND)
```
jbonofre: `command` has no `property` binding, unlike `image` below. The new docs list both parameters as equivalently configurable, but `-Dcommand=...` will silently be ignored (only setting via `<configuration>` in the POM works).


```suggestion
    @Parameter(defaultValue = DEFAULT_COMMAND, property = "command")
```

jbonofre (changes requested):
Maybe also worth to double check `examples/karaf-docker-example` if it needs an update.

Alwaysgaurav1:
Thanks for the review, @jbonofre! Good catch on both points.

I've pushed an update addressing this feedback:
1. **CLI property binding for `command`**: Added `property = "command"` to the `@Parameter` annotation in `DockerfileMojo.java` so that `-Dcommand=...` works as expected.
2. **Updated `examples/karaf-docker-example`**: Both `static-dist` and `dynamic-dist` specify `<javase>17</javase>`, so I configured `<image>eclipse-temurin:17-jre</image>` in their `dockerfile` executions to align with Java 17 (preventing the default Java 11 JRE from being used). I also updated the legacy `FROM openjdk:8-jre` sample traces in the example's `README.md`.

jbonofre:
@Alwaysgaurav1 great! Thanks a lot! I'm doing a new pass 😀

Alwaysgaurav1:
@jbonofre okayy i'm waiting.

tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/DockerfileMojo.java
```diff
@@ -31,29 +31,46 @@
...
         File dockerFile = new File(destDir, "Dockerfile");
         try {
             StringBuilder buffer = new StringBuilder();
-            buffer.append("FROM eclipse-temurin:11-jre").append("\n");
+            buffer.append("FROM ").append(baseImage).append("\n");
             buffer.append("ENV KARAF_INSTALL_PATH /opt").append("\n");
             buffer.append("ENV KARAF_HOME $KARAF_INSTALL_PATH/apache-karaf").append("\n");
             buffer.append("ENV KARAF_EXEC exec").append("\n");
             buffer.append("ENV PATH $PATH:$KARAF_HOME/bin").append("\n");
             buffer.append("COPY ").append(assembly.getName()).append(" $KARAF_HOME").append("\n");
```
jbonofre, line 71: `assembly.getName()` is spliced into this `COPY` line with no validation, unlike `image`/`command` below.
Since this PR's whole point is to stop unvalidated strings reaching raw Dockerfile lines, a filename containing a literal newline (legal on Unix) would let this inject an arbitrary extra Dockerfile instruction.
I propose the same guard applied here. It's your call if you are happy to do it, else I will merge as is.

tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/DockerfileMojo.java
```diff
@@ -31,29 +31,46 @@
...
+    private String image;
+
     @Override
     public void execute() throws MojoExecutionException, MojoFailureException {
         getLog().info("Creating Dockerfile");
+
+        String baseImage = (image == null || image.trim().isEmpty()) ? DEFAULT_IMAGE : image.trim();
+        String cmd = (command == null || command.trim().isEmpty()) ? DEFAULT_COMMAND : command.trim();
+
+        if (baseImage.contains("\n") || baseImage.contains("\r")) {
```
jbonofre, line 56: This only blocks `\n`/`\r`. A same line value like `--platform=linux/386 alpine AS build` has no newline but still changes what `FROM` does (adds a platform override + build stage) once written out.
The newline-only blacklist doesn't actually close the gap this PR claims to close.

tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/DockerfileMojo.java
```diff
@@ -31,29 +31,46 @@
...
+            throw new MojoExecutionException("Invalid image: base image cannot contain newline characters");
+        }
+        if (cmd.contains("\n") || cmd.contains("\r")) {
```
jbonofre, line 59: Same gap as `image` above: `cmd` is only checked for `\n`/`\r`, not validated as well-formed JSON exec-form. A typo like `karaf run` (missing brackets/quotes) passes this check and silently becomes shell-form `CMD karaf run`, changing PID 1/signal-forwarding behavior with no warning.

tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/DockerfileMojo.java
Alwaysgaurav1: Hi @jbonofre, thanks for the review!

Regarding the validation: the newline checks were added specifically to prevent multi-line Dockerfile directive injection while retaining full flexibility for valid single-line Docker options (such as --platform=... on FROM and shell-form on CMD).

Since you mentioned you are comfortable merging as-is, that works great for me! Thank you for reviewing and guiding this PR.

jbonofre (approved):
Thanks! It works for me too 😄 
I will merge as soon as CI is green.

PR #2851, opened 2026-09-06
Remove deprecated Cookie calls in ProxyServlet

holgerfriedrich:
`Cookie.setComment(String)` and `Cookie.setVersion(int)` are deprecated for removal in Jakarta Servlet 6.1 and specified as no-ops: `setComment` is ignored and `getComment()` returns null, `setVersion` is ignored and `getVersion()` returns 0.

See documentation
https://jakarta.ee/specifications/servlet/6.1/apidocs/jakarta.servlet/jakarta/servlet/http/cookie

> With the adoption of support for RFC 6265, this method should no longer be used.
>
> If called, this method has no effect.


oscerd (approved):

jbonofre:
As original author of the `ProxyServlet` (😄 ) I will review. Thanks!

jbonofre (approved):

PR #2852, opened 2026-09-06
Install the SIGHUP dump handler reflectively

holgerfriedrich:
This avoids warnings during compilation which are printed in every CI summary.
sun.misc.Signal and SignalHandler are internal proprietary APIs. The approach is taken over from Main.java.

### DIFF
--- a/diagnostic/boot/src/main/java/org/apache/karaf/diagnostic/core/internal/Activator.java
+++ b/diagnostic/boot/src/main/java/org/apache/karaf/diagnostic/core/internal/Activator.java
@@ -16,28 +16,24 @@
  */
 package org.apache.karaf.diagnostic.core.internal;
 
-import java.io.Closeable;
-
 import org.osgi.framework.BundleActivator;
 import org.osgi.framework.BundleContext;
 
 public class Activator implements BundleActivator {
-    Closeable dumpHandler;
+    DumpHandler dumpHandler;
 
     public void start(BundleContext context) throws Exception {
         if (!isWindows()) {
-            ClassLoader cl = this.getClass().getClassLoader();
             try {
-                Class<?> dumpHandlerClazz = cl.loadClass("org.apache.karaf.diagnostic.core.internal.DumpHandler");
-                dumpHandler = (Closeable) dumpHandlerClazz.getConstructor(BundleContext.class).newInstance(context);
-            } catch (Throwable e) {
-                // Will happen if sun.misc.SignalHandler is not available
+                dumpHandler = new DumpHandler(context);
+            } catch (Exception e) {
+                // Will happen if sun.misc.Signal is not available
             }
         }
     }
 
     public void stop(BundleContext context) throws Exception {
-        if (dumpHandler != null && !isWindows()) {
+        if (dumpHandler != null) {
             dumpHandler.close();
         }
     }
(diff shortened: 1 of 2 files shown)


jbonofre:
It sounds interesting, I will do the review.

diagnostic/boot/src/main/java/org/apache/karaf/diagnostic/core/internal/Activator.java
```diff
@@ -16,28 +16,24 @@
...
     public void start(BundleContext context) throws Exception {
         if (!isWindows()) {
-            ClassLoader cl = this.getClass().getClassLoader();
             try {
-                Class<?> dumpHandlerClazz = cl.loadClass("org.apache.karaf.diagnostic.core.internal.DumpHandler");
-                dumpHandler = (Closeable) dumpHandlerClazz.getConstructor(BundleContext.class).newInstance(context);
-            } catch (Throwable e) {
-                // Will happen if sun.misc.SignalHandler is not available
+                dumpHandler = new DumpHandler(context);
+            } catch (Exception e) {
```
jbonofre: This narrows the previous `catch (Throwable e)` to `Exception`. The reflective `DumpHandler` constructor can fail with `Error` subtypes rather than exceptions (`ExceptionInInitializerError`, `NoClassDefFoundError`, `LinkageError`). On JVMs where `sun.misc.Signal` resolves as a name but can't be initialized or linked (GraalVM native image, `-Xrs`, restricted/embedded JVMs).

Those would now escape `start()` and fail activation of diagnostic bundle, which is the exact "Signal not available" case this catch is meant to tolerate. 

Please keep `catch (Throwable e)` (or `catch Exception | LinkageError`).

diagnostic/boot/src/main/java/org/apache/karaf/diagnostic/core/internal/DumpHandler.java
```diff
@@ -27,20 +29,41 @@
...
-        previous = sun.misc.Signal.handle(new Signal(SIGNAL), this);
+
+        final Class<?> signalClass = Class.forName("sun.misc.Signal");
+        final Class<?> signalHandlerClass = Class.forName("sun.misc.SignalHandler");
+
+        Object signalHandler = Proxy.newProxyInstance(getClass().getClassLoader(),
+            new Class<?>[] {
+                signalHandlerClass
+            },
+                (proxy, method, args) -> {
```
jbonofre, line 55: The proxy ignores `method`, so every call dispatched to this handler (including `Object.equals`/`hashCode`/`toString`) triggers a full diagnostic dump zip in the working directory. 

I suggest the following guarding:

```java
(proxy, method, args) -> {
  if ("handle".equals(method.getName())) {
    handle();
    return null;
  }
  return method.invoke(this, args); // Object methods
}
```

diagnostic/boot/src/main/java/org/apache/karaf/diagnostic/core/internal/DumpHandler.java
```diff
@@ -27,20 +29,41 @@
...
+        );
+
+        handleMethod = signalClass.getMethod("handle", signalClass, signalHandlerClass);
+        signal = signalClass.getConstructor(String.class).newInstance(SIGNAL);
+        previous = handleMethod.invoke(null, signal, signalHandler);
     }
-    
-    public void handle(Signal signal) {
+
+    private void handle() {
```
jbonofre, line 74: `Dump.dump` (collect and zip everything) runs synchronously on the JVM signal-dispatch thread here. `Main.registerSignalHandler` deliberately offloads its handler body to `new Thread(...)`.

This is pre-existing behaviour, but since the method being rewritten anyway it would be a good moment to match `Main` and run the dump on a short-lived thread.

jbonofre (changes requested):

holgerfriedrich:
changed...

jbonofre (approved):

PR #2870, opened 2026-09-09
Avoid the non-varargs call warning in AsmProxyFactory

holgerfriedrich:
Object.class.getConstructor(null) is a non-varargs call of a varargs method, which javac reports in every CI summary. The lookup could not return anything but the no-arg constructor of Object, so its descriptor was always the "()V" the variable was already initialized with, and the NoSuchMethodException was never thrown. Drop the whole construct.

The constant descriptor means the proxied class must have a no-arg constructor, which nothing verified: the generated INVOKESPECIAL only failed once the proxy was instantiated. Check it upfront and report it as an IllegalArgumentException instead.

This PR was created with help of Claude Code.

### DIFF
--- a/services/interceptor/impl/src/main/java/org/apache/karaf/service/interceptor/impl/runtime/proxy/AsmProxyFactory.java
+++ b/services/interceptor/impl/src/main/java/org/apache/karaf/service/interceptor/impl/runtime/proxy/AsmProxyFactory.java
@@ -121,22 +121,7 @@ public class AsmProxyFactory {
             parentClassFileName = Type.getInternalName(Object.class);
         } else {
             parentClassFileName = classFileName;
-            // without these checks the generated INVOKESPECIAL would only fail once the proxy is
-            // instantiated. The proxy is defined by its own class loader, so it lands in a different
-            // runtime package than the proxied class: only a public or protected constructor is
-            // reachable from it, a package-private one is not.
-            final Constructor<?> superCt;
-            try {
-                superCt = classToProxy.getDeclaredConstructor();
-            } catch (final NoSuchMethodException nsme) {
-                throw new IllegalArgumentException("Cannot proxy " + classToProxy.getName()
-                        + ", it has no no-arg constructor", nsme);
-            }
-            final int modifiers = superCt.getModifiers();
-            if (!Modifier.isPublic(modifiers) && !Modifier.isProtected(modifiers)) {
-                throw new IllegalArgumentException("Cannot proxy " + classToProxy.getName()
-                        + ", its no-arg constructor is not accessible from the generated proxy");
-            }
+            checkProxyable(classToProxy);
         }
         final String descriptor = "()V";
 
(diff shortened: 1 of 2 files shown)


holgerfriedrich:
Screenshot_20260909-203631

Warnings are gone.

services/interceptor/impl/src/main/java/org/apache/karaf/service/interceptor/impl/runtime/proxy/AsmProxyFactory.java
```diff
@@ -114,19 +113,22 @@ private boolean hasSameSignature(Method a, Method b) {
...
-                descriptor = Type.getConstructorDescriptor(superDefaultCt);
+        // the proxy extends the proxied class, or Object when proxying an interface; either way
+        // the super constructor it invokes is the no-arg one
+        final String parentClassFileName;
+        if (classToProxy.isInterface()) {
+            parentClassFileName = Type.getInternalName(Object.class);
+        } else {
+            parentClassFileName = classFileName;
+            try {
+                classToProxy.getDeclaredConstructor();
```
jbonofre: `getDeclaredConstructor()` also matches a private or package-private no-arg constructor, but the `INVOKESPECIAL` we generate below can't call either one. So far those classes the check passes and we still fail late, which is exactly what this guard is meant to prevent.

Package-private doesn't help here either: the proxy is define by a fresh `ProxyFactory.ProxyClassLoader`, so it ends up in a different runtime package than the proxied class even though the package name matches.

`protected` does work (super-constructor access from a direct subclass is legal across packages), so the condition we actually want is "public or protected", not "declared". Concrete case that regresses today: a service impl registered via `bundleContext.registerService(...)` with a package-private no-arg constructor (a pretty normal idiom) gets an `IllegalStateException`/`IllegalAccessError` instead of the clear `IllegalArgumentException`.

The simplest fix is `getConstructor()`, which is public-only and also avoids needing `accessDeclaredMembers` under a security manager:

```java
classToProxy.getConstructor();
```

If you want to keep `protected` proxyable, keep `getDeclaredConstructor()` and reject on the modifiers:


```java
final Constructor<?> superCt = classToProxy.getDeclaredConstructor();
final int modifiers = superCt.getModifiers();
if (!Modifier.isPublic(modifiers) && !Modifier.isProtected(modifiers)) {
  throw new IllegalArgumentException("Cannot proxy " + classToProxy.getName() + ", its no-arg constructor is not accessible from the generated proxy");
}
```

Maybe worth a test alongside `proxyWithoutNoArgConstructor` covering a class whose only no-arg constructor is package-private.

jbonofre (changes requested):

services/interceptor/impl/src/main/java/org/apache/karaf/service/interceptor/impl/runtime/proxy/AsmProxyFactory.java
```diff
@@ -114,19 +114,31 @@ private boolean hasSameSignature(Method a, Method b) {
...
+            parentClassFileName = Type.getInternalName(Object.class);
+        } else {
+            parentClassFileName = classFileName;
+            // without these checks the generated INVOKESPECIAL would only fail once the proxy is
+            // instantiated. The proxy is defined by its own class loader, so it lands in a different
+            // runtime package than the proxied class: only a public or protected constructor is
+            // reachable from it, a package-private one is not.
+            final Constructor<?> superCt;
+            try {
+                superCt = classToProxy.getDeclaredConstructor();
```
jbonofre: `getDeclaredConstructor()` forces the JVM to resolve **all** declared constructors of `classToProxy`, not just the no-arg one: previously this branch never reflected on constructors at all.

If `classToProxy` has an unrelated constructor whose parameter type lives in an optional/unresolvable OSGi bundle, this now throws an uncaught `NoClassDefFoundError` where proxying used to just work.

That's a regression risk in exactly the split-classpath environment Karaf runs in. Worth maybe a test with such a class before merging. Thoughts?

services/interceptor/impl/src/main/java/org/apache/karaf/service/interceptor/impl/runtime/proxy/AsmProxyFactory.java
```diff
@@ -114,19 +114,31 @@ private boolean hasSameSignature(Method a, Method b) {
...
+            } catch (final NoSuchMethodException nsme) {
+                throw new IllegalArgumentException("Cannot proxy " + classToProxy.getName()
+                        + ", it has no no-arg constructor", nsme);
+            }
+            final int modifiers = superCt.getModifiers();
+            if (!Modifier.isPublic(modifiers) && !Modifier.isProtected(modifiers)) {
```
jbonofre: This only checks the constructor own modifiers, not whether `classToProxy` itself is accessible to a class defined by a different `ClassLoader`. A package-private class (common for OSGi impl classes) with a public/protected no-arg constructor passes this check, but the proxy (defined by its own `ProxyClassLoader`, hence a different runtime package even with an identical package name) still can't legally extend it.

That fails later with an uncaught `IllegalAccessError` at class-link time, which is exactly the failure mode this PR set out to prevent.

I suggest also checking `Modifier.isPublic(classToProxy.getModifiers())` here. Thoughts?

jbonofre (commented):
Sorry to be a pain, I have a couple of questions for you here 😄

services/interceptor/impl/src/main/java/org/apache/karaf/service/interceptor/impl/runtime/proxy/AsmProxyFactory.java
holgerfriedrich: Yes, done. Extended the test cases to cover it as well.

services/interceptor/impl/src/main/java/org/apache/karaf/service/interceptor/impl/runtime/proxy/AsmProxyFactory.java
holgerfriedrich: let's catch `LinkageError`, this is what fails with the new tests.
Unfortunately it is not a declared exception of those functions.

holgerfriedrich:
Yes, this turned out to be a bigger thing than initially expected. :see_no_evil: 
Probably I should spent some more time on the J25 side to get ready for 4.5..... I have not done much testing on top of the recently merged changes.

jbonofre (approved):
I pushed directly my suggestion to move forward on the PR and it looks good to me now.
I wil merge as soon as CI is green.

PR #2892, opened 2026-09-12
Restrict WebConsole plugin getResource() to /res/ paths

jbonofre:
## Summary
- `FeaturesPlugin`, `GogoPlugin`, `HttpPlugin` and `InstancePlugin` override `getResource()` and hand the request path directly to `classLoader.getResource()` with no restriction.
- Felix's own `AbstractServlet.getResource()` (org.apache.felix.webconsole) only spools paths whose remainder starts with `/res/`; these four Karaf plugins dropped that check.
- Restores the `/res/` prefix check in all four plugins so only each plugin's own bundled static resources (JS/CSS/images) can be served through this path, matching upstream Felix behavior.

## Test plan
- [x] `mvn -pl webconsole/features,webconsole/gogo,webconsole/http,webconsole/instance -am compile` succeeds
- [x] `mvn -pl webconsole/features,webconsole/gogo,webconsole/http,webconsole/instance -am test` succeeds (existing `InstancePluginTest` passes)
- [ ] Manual verification that `/system/console/features/res/...`, `/system/console/gogo/res/...`, `/system/console/http/res/...`, `/system/console/instance/res/...` still resolve their JS/CSS/image resources correctly in a running instance

### DIFF
--- a/webconsole/gogo/src/main/java/org/apache/karaf/webconsole/gogo/GogoPlugin.java
+++ b/webconsole/gogo/src/main/java/org/apache/karaf/webconsole/gogo/GogoPlugin.java
@@ -97,7 +97,7 @@ public class GogoPlugin extends AbstractServlet {
     @Override
     protected URL getResource(String path) {
         path = path.substring(NAME.length() + 1);
-        if (path == null || path.isEmpty()) {
+        if (path == null || path.isEmpty() || !path.startsWith("/res/")) {
             return null;
         }
         URL url = this.getClass().getClassLoader().getResource(path);
--- a/webconsole/features/src/main/java/org/apache/karaf/webconsole/features/FeaturesPlugin.java
+++ b/webconsole/features/src/main/java/org/apache/karaf/webconsole/features/FeaturesPlugin.java
@@ -132,7 +132,7 @@ public class FeaturesPlugin extends AbstractServlet {
     @Override
     protected URL getResource(String path) {
         path = path.substring(NAME.length() + 1);
-        if (path == null || path.isEmpty()) {
+        if (path == null || path.isEmpty() || !path.startsWith("/res/")) {
             return null;
         }
         URL url = this.classLoader.getResource(path);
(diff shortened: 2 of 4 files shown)


webconsole/gogo/src/main/java/org/apache/karaf/webconsole/gogo/GogoPlugin.java
```diff
@@ -97,7 +97,7 @@ public void renderContent(HttpServletRequest request, HttpServletResponse respon
     @Override
     protected URL getResource(String path) {
         path = path.substring(NAME.length() + 1);
-        if (path == null || path.isEmpty()) {
+        if (path == null || path.isEmpty() || !path.startsWith("/res/")) {
```
CptBartender: `path == null` nullcheck might be superfluous, as the `String#substring` call above would either throw an NPE if `path` is `null`, or return a valid (potentially empty) string.

jbonofre: Thanks! Actually the path null guard should be before substring. Let me fix that (it was like this before, not related to this PR).

holgerfriedrich:
Maybe we should use something like `URI.create(path).normalize().getPath()` to avoid path traversal like `/res/../forbidden_directory/`?
Hopefully Jetty does this anyway, including other character escapings, but I am not sure.

webconsole/gogo/src/main/java/org/apache/karaf/webconsole/gogo/GogoPlugin.java
jbonofre: Done

jbonofre:
> Maybe we should use something like `URI.create(path).normalize().getPath()` to avoid path traversal like `/res/../forbidden_directory/`? Hopefully Jetty does this anyway, including other character escapings, but I am not sure.

@holgerfriedrich good point, thanks!

PR #2893, opened 2026-09-12
[4.4.x] Upgrade org.eclipse.osgi from 3.18.0 to 3.18.600

### LINKED ISSUE #2731: Critical CVE in equinox 3.18.0 in karaf.assemblies.features
Our repository dependency blocker has flagged and quarantained the following artifact:
https://mvnrepository.com/artifact/org.apache.karaf.assemblies.features/features/4.4.11

The underlying cause is most likely the following critical issue (9.8)
https://guide.sonatype.com/vulnerability/CVE-2023-54342
in equinox 3.18.0.
Could it be possible to upgrade this version to 3.18.600? I believe it is fully compatible, it would be great if it could be fixed backwards also in at least version 4.4.11.

holgerfriedrich:
This PR included the changes to handle comment in  #2303.

This may fix #2731, but not necessarily, as the CVE has no version range specified for the 3.18.x series, or?

@jbonofre I have used 3.18.600 for maximum compatibility, but maybe we should use a later version. Please advise.

### DIFF
--- a/main/src/main/java/org/apache/karaf/main/Main.java
+++ b/main/src/main/java/org/apache/karaf/main/Main.java
@@ -537,7 +537,14 @@ public class Main {
         if (factoryClass == null) {
             InputStream is = classLoader.getResourceAsStream("META-INF/services/" + FrameworkFactory.class.getName());
             BufferedReader br = new BufferedReader(new InputStreamReader(is, StandardCharsets.UTF_8));
-            factoryClass = br.readLine();
+            String line;
+            while ((line = br.readLine()) != null) {
+                line = line.trim();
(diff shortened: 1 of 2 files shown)


jbonofre:
@holgerfriedrich It's fine to update to Equiinox 3.18.600, but I would like to clarify the CVE reason.

The CVE describes RCE via a fork console command reachable over an unauthorized telnet connection to Equinox's console interface.
Karaf doesn't use osgi.console from Equinox as we use our own SSH server/shell, not Equinox's built-in console, so we are not vulnerable.

So, upgrading to make the scanner happy (people just see that without checking if it's actually vulnerable 😄 ),but we are not fixing the CVE as we are not impacted.

Issue #2741, opened 2026-06-23
upgrade major jdk version makes Karaf unresponsive to filesystem changes


gerbenasd:
(a minor upgrade did not show this problem to me, bu might also be problematic)

Simplest example: changes to file org.ops4j.pax.logging.cfg

This is also seen with changes to files in the deploy directory (using camel-blueprint).
This might be problematic in other situations as well.

Way to resolve problem is to
 - stop karaf
 - remove data directory
 - start karaf  (changes to org.ops4j.pax.logging.cfg are reflected correctly)
 (this requires rebuilding the karaf instance to you needs)

### EXAMPLE

Unpack a fresh download (apache-karaf-4.4.11.zip)
Start karaf.bat using JAVA_HOME=C:\opt\Java\jdk-11.0.27+6
Changes to org.ops4j.pax.logging.cfg are tracked, e.g. changing filename repeatedly:
	log4j2.appender.rolling.fileName = ${karaf.log}/karaf.log
	log4j2.appender.rolling.fileName = ${karaf.log}/karaf1.log
	log4j2.appender.rolling.fileName = ${karaf.log}/karaf2.log

Stop and start using JAVA_HOME=C:\opt\Java\jdk-17.0.14+7
This already creates unresponsiveness. Even reverting back to previous jdk will not help anymore.

changing filename is not detected anymore (log:tail will show nothing about that change)
	log4j2.appender.rolling.fileName = ${karaf.log}/karaf3.log

the logfile will remain karaf2.log

###  UNDER JDK 11
C:\opt\apache-karaf-4.4.11\bin>set JAVA
JAVA_HOME=C:\opt\Java\jdk-11.0.27+6

C:\opt\apache-karaf-4.4.11\bin>time
The current time is: 13:48:27,43
Enter the new time:

C:\opt\apache-karaf-4.4.11\bin>karaf.bat
karaf.bat: KARAF_LOG doesn't exist: "C:\opt\apache-karaf-4.4.11\bin\..\data\log"
karaf.bat: Creating "C:\opt\apache-karaf-4.4.11\bin\..\data\log"
        __ __                  ____
       / //_/____ __________ _/ __/
      / ,<  / __ `/ ___/ __ `/ /_
     / /| |/ /_/ / /  / /_/ / __/
    /_/ |_|\__,_/_/   \__,_/_/

  Apache Karaf (4.4.11)

Hit '<tab>' for a list of available commands
and '[cmd] --help' for help on a specific command.
[...]

PR #2899, opened 2026-09-13
[karaf-4.4.x] GH-2741: Clean the OSGi bundle cache on JDK version change

jbonofre:
## Summary

Backport of #2898 to `karaf-4.4.x`.

After a major JDK version change (e.g. JDK 11 -> 17), Karaf can become unresponsive to `etc/*.cfg` changes and deploy directory drops (`log:tail` shows nothing about the change). This was tracked down to a stale bundle wiring persisted in `data/cache`: after restarting under the new JDK, the File Install bundle's cached wiring was missing wires to `org.osgi.service.cm` (Configuration Admin) and `org.osgi.service.log` (Log Service), so config/deploy directory monitoring silently stops working. Reverting to the previous JDK does not self-heal it, since the corrupted state is on disk, not JDK-dependent.

This change makes Karaf detect the situation and clean up automatically:

- `ConfigProperties` now records the JDK specification version (`java.specification.version`) used on the last start in `data/jdk_version`.
- On startup, if the recorded version differs from the current one, the OSGi bundle cache (`org.osgi.framework.storage`, normally `data/cache`) is cleaned before the framework is created, forcing a full re-resolution instead of reusing stale wiring.
- A normal restart on the same JDK (including patch upgrades, since only the specification version is compared) is unaffected.
- The behavior can be disabled with `-Dkaraf.clean.cache.on.jdk.change=false` for anyone who wants to manage this themselves.

Cherry-picked cleanly from `main` (one trivial context difference in `ConfigProperties.java`, unrelated to this change).

## Test plan

- [x] `mvn -pl util,main -am install -DskipTests` then `mvn -pl main test -Dtest=UtilsTest` passes on `karaf-4.4.x`.

### DIFF
--- a/main/src/main/java/org/apache/karaf/main/util/Utils.java
+++ b/main/src/main/java/org/apache/karaf/main/util/Utils.java
@@ -24,9 +24,47 @@ import java.io.IOException;
 import java.net.JarURLConnection;
 import java.net.URI;
 import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.util.StringTokenizer;
 
 public class Utils {
+
+    /**
+     * Name of the marker file, stored in the Karaf data directory, recording the
+     * specification version of the JDK that was used for the last start of the instance.
+     */
+    public static final String JDK_VERSION_MARKER_FILE = "jdk_version";
+
+    /**
+     * Detects a JDK (specification) version change since the last start of this instance and,
+     * when detected, cleans the OSGi bundle cache before the framework is created.
+     * <p>
+     * The bundle cache can persist a bundle wiring that was resolved against a different JDK
+     * (different major version exposes a different set of packages to the system bundle). Reusing
+     * such a stale wiring after a JDK upgrade/downgrade can silently leave core bundles (e.g.
+     * Configuration Admin, File Install) without expected wires, without any error being reported.
+     *
+     * @param karafData the Karaf data directory, where the marker file is stored
+     * @param cacheDirectory the OSGi framework bundle cache directory
+     * @throws IOException if the marker file can't be read/written, or the cache can't be cleaned
+     */
+    public static void cleanCacheOnJdkChange(File karafData, File cacheDirectory) throws IOException {
+        String currentVersion = System.getProperty("java.specification.version");
+        File marker = new File(karafData, JDK_VERSION_MARKER_FILE);
+        String previousVersion = null;
+        if (marker.exists()) {
+            previousVersion = new String(Files.readAllBytes(marker.toPath()), StandardCharsets.UTF_8).trim();
+        }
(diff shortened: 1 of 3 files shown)


main/src/main/java/org/apache/karaf/main/util/Utils.java
```diff
@@ -24,10 +24,48 @@
...
+        }
+        if (previousVersion != null && !previousVersion.isEmpty() && !previousVersion.equals(currentVersion)) {
+            System.err.println("WARN: JDK version change detected since the last start (" + previousVersion + " -> " + currentVersion
+                    + "). Cleaning the OSGi bundle cache (" + cacheDirectory + ") to avoid reusing a bundle wiring resolved against a different JDK.");
+            deleteDirectory(cacheDirectory);
+            cacheDirectory.mkdirs();
+        }
+        karafData.mkdirs();
+        Files.write(marker.toPath(), currentVersion.getBytes(StandardCharsets.UTF_8));
+    }
```
CptBartender: The proposed code silently ignores any unrecognized characters within the file. Also, using `Files#readString` is marginally better memory-wise, so I suggest the following change:

```suggestion
    public static void cleanCacheOnJdkChange(File karafData, File cacheDirectory) throws IOException {
        String currentVersion = System.getProperty("java.specification.version");
        String previousVersion = getPreviousVersionFromFile(karafData);
        if (previousVersion != null && !previousVersion.equals(currentVersion)) {
            System.err.println("WARN: JDK version change detected since the last start ('" + previousVersion + "' -> '" + currentVersion
                    + "'). Cleaning the OSGi bundle cache (" + cacheDirectory + ") to avoid reusing a bundle wiring resolved against a different JDK.");
            deleteDirectory(cacheDirectory);
            cacheDirectory.mkdirs();
        }
        karafData.mkdirs();
        Files.write(marker.toPath(), currentVersion.getBytes(StandardCharsets.UTF_8));
    }
    
    /* 
     * Returns the version from file if successfully retrieved, static "invalid" string if file contains
     * invalid characters, or null for when the file does not exist
     * /
    private static String getPreviousVersionFromFile(File karafData) throws IOException {
        File marker = new File(karafData, JDK_VERSION_MARKER_FILE);
        if (marker.exists()) {
            try {
                return Files.readString(marker.toPath(), StandardCharsets.UTF_8).trim();
            } catch (CharacterCodingException e) {
                System.err.println("WARN: previous JDK version contains invalid characters");
                return "invalid";
            }
        } else {
            return null;
        }
    }
```

`CharacterCodingException` is a descendant of `IOException` and is declared by the method that actually handles string creation within `Files#readString`

I chose to add a non-javadoc comment to the extracted private method, assuming it's enough in this case.

I also added wrapping of the logged JDK versions with `'apostrophes'` so as to clarify in case the previous version is ` `

PR #2911, opened 2026-09-15
Fixed various IntelliJ warnings in BaseActivator

CptBartender:
- Added type parameters to generics
- Changed the class so that it explicitly implements ManagedService instead of sneakily registering itself as such Extracted the string splitting pattern to a static field so that the regex is compiled only once
- Changed getInterfaceNames so that it returns distinct classes in case any implemented interfaces extend the same superinterface
- Changed trackers map to use Class objects as keys instead of String names
- Intermediately bumped required Java version to 17 so that pattern matching for `instanceof` can be used - eventually it'll be bumped in https://github.com/apache/karaf/pull/2214, I presume


util/src/main/java/org/apache/karaf/util/tracker/BaseActivator.java
```diff
@@ -320,40 +328,39 @@ public void run() {
...
     }
 
     protected void trackService(String className, String filter) throws InvalidSyntaxException {
-        if (!trackers.containsKey(className)) {
-            SingleServiceTracker tracker = new SingleServiceTracker<>(bundleContext, className, filter, (u, v) -> reconfigure());
-            tracker.open();
-            trackers.put(className, tracker);
-        }
+      try {
+        Class<?> clazz = Class.forName(className);
```
jbonofre, line 359: This changes `trackService(String, String)` from pure name-based tracking to eager class resolution via `Class.forName`.

I have two concerns here:
1. This method is `protected`. A subclass calling it with a class name not importable by its own bundle will now get a silently-dropped tracker (just a WARN log) instead of the name-based OSGi filter tracking this overload exists to provide.
2. Only `ClassNotFoundException` is caught. A bad static initializer or a missing transitively-referenced class raises `ExceptionInInitiliazerError`/`NoClassDefFoundError`, which isn't caught here and will propagate out of `doOpen()`/`start(BundleContext)`, aborting the whole bundle's activation. The old code could never throw from this path.

util/src/main/java/org/apache/karaf/util/tracker/BaseActivator.java
```diff
@@ -86,8 +96,9 @@ public void start(BundleContext context) throws Exception {
         doOpen();
         scheduled.set(false);
         if (managedServiceRegistration == null
-                && trackers.values().stream()
-                    .allMatch(t -> t.getService() != null)) {
+            && trackers.values().stream()
+            .map(SingleServiceTracker::getService)
+            .allMatch(Objects::nonNull)) {
```
jbonofre, line 101: `allMatch(...)` on `trackers.values()` is vacuously `true` when `trackers` is empty. If a required service's class fails to resolve it `trackService(String, String)`, it's never added to `trackers` at all, so this check now incorrectly reports "ready" instead of waiting.

`doStart()` then calls `getTrackedServices(RequiredInterface.class)`, which throws `IllegalStateException("Service not tracked for class ...)`: a confusing crash disconnected from the real (silently logged) root cause.

util/src/main/java/org/apache/karaf/util/tracker/BaseActivator.java
```diff
@@ -364,15 +371,17 @@ protected void trackService(String className, String filter) throws InvalidSynta
      * @return The actual tracker service object.
      */
     protected <T> T getTrackedService(Class<T> clazz) {
-        SingleServiceTracker tracker = trackers.get(clazz.getName());
+        @SuppressWarnings("unchecked")
```
jbonofre, line 374: The runtime type check on the returned service was dropped. Previously `clazz.cast(tracker.getService())` threw an immediate, clearly-attributed `ClassCastException` if the tracked object wasn't actually assignable to `clazz` (e.g. duplicate class definitions across bundle classloaders). Now it's an unchecked cast on the tracker itself, and `tracker.getService()` is returned with no check: a type mismatch surfaces later (if at all) as a confusing `ClassCastException` far from the real cause, or not at all if passed through untyped code.

util/src/main/java/org/apache/karaf/util/tracker/BaseActivator.java
```diff
@@ -35,29 +37,37 @@
...
+import org.osgi.framework.ServiceReference;
+import org.osgi.framework.ServiceRegistration;
 import org.osgi.service.cm.Configuration;
 import org.osgi.service.cm.ConfigurationAdmin;
+import org.osgi.service.cm.ManagedService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class BaseActivator implements BundleActivator, Runnable, ThreadFactory {
+public class BaseActivator implements BundleActivator, ManagedService, Runnable, ThreadFactory {
```
jbonofre, line 55: Making `BaseActivator` directly `implement ManagedService` turns `org.osgi.service.cm` into a hard compile/link-time dependency for every bundle embedding this class, not just ones that call `manage(pid)`.

The ~ 20 pom.xml additions in this PR cover in-tree callers, but this is a compatibility-breaking change for any downstream/third-party Karaf-based budle that extends `BaseActivator` (including Karaf subprojects like Cellar or Decanter) without Configuration Admin on its classpath: it would now fail to load the class at all (`NoClassDefFoundError`), even if it never calls `manage()`.

I think it's worth calling out explicitly since `BaseActivator` is public API.

jbonofre (changes requested):

PR #2920, opened 2026-09-19
[4.4.x] bump net.bytebuddy:byte-buddy to 1.17.8

holgerfriedrich:
We had a version drift between pom and feature files. Dependabot only upgraded pom.
1.17.8 aligns with hibernate6 and is latest version of the 1.17.x branch.
Note that hibernate JPA feature still pulls 1.12.9 (which is not even latest of 1.12.x branch). This is kept.

> Introduce a property to ensure feature and pom keeps in sync where needed. Downgrade to 1.17.x for 4.4.x branch.


holgerfriedrich:
@jbonofre the question is still, if we should use 1.18.14.
I tend to say "we should not for 4.4.x", so I created this one on 1.17.8.
I have no strong opinion and will change that if you like.

jbonofre:
@holgerfriedrich agree, that's cleaner. Let's stay on 1.17.8 for 4.4.x.

jbonofre (approved):

PR #2922, opened 2026-09-19
[test] Stop using SSH keys from .ssh folder

holgerfriedrich:
It seems we pull in keys from local .ssh folder in tests.


jbonofre:
@holgerfriedrich can you provide some rationale? It was intentional (as we can use key auth with "regular" OpenSSH keys).

holgerfriedrich:
Sure.
I had an SSH config on my machine including a password protected key which matches any host:
~~~
Host *
    IdentityFile ~/.ssh/password_protected_ed25519_key
    IdentitiesOnly yes
~~~

This leads to SSH itests failing locally. Test result depends on developer machine.

It seems only the client side is affected, as it is set to pull in the config from `.ssh` folder and then propagates any failure opening the key to the calling test.

The change should cover the path for `IdentitiesOnly` settings, see here:
https://github.com/apache/mina-sshd/blob/sshd-2.19.0/sshd-core/src/main/java/org/apache/sshd/client/SshClient.java#L815-L833

It disables the default mechanism for implicit key discovery only. Keys registered explicitly are unaffected.

jbonofre:
@holgerfriedrich yes, it's what I thought (it's client side). I'm fine with the change. Thanks for confirming!

jbonofre (approved):

PR #2929, opened 2026-09-21
Fix webconsole over-inclusion

rovarga:
The ",**" ends up expading to include all dependencies, so drop that.


holgerfriedrich:
Reference: follow-up to https://github.com/apache/karaf/pull/2729

holgerfriedrich (approved):
Removes the additional files, and the modified class file still there 👍 
LGTM.

jbonofre (approved):

